Issue #1, opened 2022-12-12
Returning a "wrapper ndarray" from the wrapper functions


rgommers:
The current version of `wrapper.py` has:
```
# 1. Q: What are the return types of wrapper functions: plain torch.Tensors or
#       wrapper ndarrays.
#    A: Tensors, apparently
```
I think the answer has to be "wrapper ndarray", otherwise the semantics change immediately. Let's see if @ezyang, @albanD, @lezcano agree.

ev-br:
If the goal of this can be formulated as "make numpy-compatible API to work with torch tensors", it seems reasonable that all functions in this namespace operate on tensors directly, no? 

More specifically, they consume anything that can be converted to a Tensor, and return Tensors.
This approach seems to match the spirit of numpy's array_like concept, let's call it `tensor_like`.
Also this makes the whole wrapping layer thinner (not a goal in itself, but still)

What am I missing?

In https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/2#pullrequestreview-1214278466 @ezyang said 

> how are you going to compose these functions into bigger ones otherwise?

The current POC functions seem to compose as is (technically, they are just calling `torch.as_tensor` on inputs.):

```
In [1]: from wrapper import log, abs

In [2]: abs(2)
Out[2]: tensor(2)

In [3]: log(abs(2))
Out[3]: tensor(0.6931)
```

Also in the current version, `asarray` converts its input to a tensor, too:

```
In [4]: from wrapper import asarray

In [5]: asarray(2)
Out[5]: tensor(2)

In [7]: log(abs(asarray([2])))
Out[7]: tensor([0.6931])
```

lezcano:
I think the point here is that if `log` had different semantics in PyTorch than in NumPy, `log(abs(2))` the same results in NumPy than in wrapped if `abs` returns a `torch.Tensor`. 

For things to be composable, if the user has to explicitly choose to use certain array semantics (perhaps via that `wrapped import`) the user must also explicitly choose when to stop using them (by manually casting the tensor back). Otherwise, there's going to be some inconsistency lurking around.

ezyang:
For function calls I agree asarray saves you, but method calls require the wrapper

ev-br:
Okay, https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/3  starts formulating the interface with ndarrays only.

PR #2, opened 2022-12-12
A dummy PR for line-by-line reviews

ev-br:
The structure is a bit ad hoc: autogenerated stuff and autogen scripts are all in autogen/, not much need to look at it; real stuff is in `torch_np/`, which is now structured like a package. Minimal testsing is in `torch_np/tests`, set the `PYTHONPATH` env variable to run them locally. 

This will all get cleaned up, eventually.


wrapper.py
```diff
@@ -0,0 +1,1278 @@
...
+        return torch.as_tensor(x).imag == 0
+    result = torch.zeros_like(x, dtype=torch.bool)
+    return result[()]
+
+
+def isreal(x):
+    if torch.is_complex(x):
+        return torch.as_tensor(x).imag == 0
+    result = torch.zeros_like(x, dtype=torch.bool)
+    return result[()]
```
lezcano, line 315: What does this syntax mean? Why not simply return `result`?

wrapper.py
```diff
@@ -0,0 +1,1278 @@
...
+        return a
+    if torch.abs(torch.imag) < tol * torch.finfo(a.dtype).eps:
+        return torch.real(a)
+    else:
+        return a
+
+
+def iscomplex(x):
+    if torch.is_complex(x):
+        return torch.as_tensor(x).imag == 0
```
lezcano, line 306:
```suggestion
        return torch.as_tensor(x).imag != 0
```

wrapper.py
```diff
@@ -0,0 +1,1278 @@
...
+
+###### reductions
+
+def argmax(a, axis=None, out=None, *, keepdims=NoValue):
+    if axis is None:
+        result = torch.argmax(a, keepdims=bool(keepdims))
+    else:
+        result = torch.argmax(a, axis, keepdims=bool(keepdims))
+    if out is not None:
+        out[...] = result
```
lezcano, line 276:
```suggestion
        out.copy_(result)
```

lezcano: Also, how are we going to do with checks (preconditions and so). Probably we don't want to care too much about them for now, but otherwise probably you need to check something like `out.shape == result.shape` or so.

This is the kind of argument that we implemented as a decorator in PrimTorch.

wrapper.py
```diff
@@ -0,0 +1,1278 @@
...
+        raise NotImplementedError
+    if M is None:
+        M = N
+    s = M - k if k >= 0 else N + k
+    z = torch.zeros(N, M, dtype=dtype)   
+    if s < 0:
+        return z
+    else:
+        ones = torch.ones(s, dtype=dtype)
+        return torch.diagonal_scatter(z, ones, offset=k)
```
lezcano, line 149: No need to compute the `s`. This can be golfed as
```python
z = torch.zeros(N, M, dtype=dtype)
z.diagonal(k).fill_(1)
```

lezcano (commented):
Perhaps it may be better to prefer calling methods on the tensors rather than `torch.` free functions to make it a little bit less PyTorch-specific? Or perhaps it doesn't matter? WDYT

Also, when we write tests for this, something we may want to test is whether the functions are returning views / copies as NumPy does.

wrapper.py
ev-br: This is a "nice" numpy-ism for getting python scalars out of zero-dim arrays (or array scalar, I never know which one is which). Observe:

```
In [61]: a = np.array(3)

In [62]: a.shape
Out[62]: ()

In [63]: a[0]
IndexError                                Traceback (most recent call last)
Input In [63], in <cell line: 1>()
----> 1 a[0]

IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed

In [64]: a[()]
Out[64]: 3
```

This particular line is straight from `In [67]: print(inspect.getsource(np.iscomplex))`

lezcano: A right. Then you may want to do `a.item()` in PyTorch. This will work for any tensor with 1 element.

ev-br: Does it then need a separate check for one or more elements?
In numpy, it's a no-op
```
In [86]: np.arange(3)[()]
Out[86]: array([0, 1, 2])
```

lezcano: It does, yes. Note that this `a[()]` is pretty much the identity function (it does not return a scalar).

I guess this is another case where you may want to add a decorator `unpack_scalar` or so with this logic, as we never do it in PyTorch but it's quite a common operation in NumPy.

rgommers:
> Also, when we write tests for this, something we may want to test is whether the functions are returning views / copies as NumPy does.

That's extremely difficult to get 100% right, or to reimplement. For some functions it will depend on, strides, memory order, and/or dimensionality (perhaps other things). For most of the time PyTorch and NumPy are going to match (both are "return a view when possible, a copy otherwise") - but I'm not sure how to deal with all the corner cases. We should probably either punt on that, or raise on a whole class of "inplace ops on maybe-views".

wrapper.py
ev-br: Am going to keep this conversation unresolved so that this point does not fall under the radar.

However @rgommers would you know the numpy plans about these things? Is this what Sebastian plans to remove or will it stay?

autogen/gen_ufuncs.py
```diff
@@ -0,0 +1,127 @@
...
+skip = {np.frexp, np.modf,    # non-standard unary ufunc signatures
+        np.isnat,
+        np.invert,      # bitwise NOT operator
+        np.spacing,     # niche, does not have a direct equivalent
+}
+
+# np functions where torch names differ
+torch_names = {np.radians : "deg2rad",
+               np.degrees : "rad2deg",
+               np.conjugate : "conj_physical",   # XXX: check
```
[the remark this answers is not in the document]
lezcano: PyTorch has a "conj bit" which returns a view of the tensor with the conjugate bit on. That one does not have a NumPy equivalent, so you won't need to use it. `conj_physical` copies the data, so it's likely the one you want.

ev-br, line 20:
```suggestion
               np.conjugate : "conj_physical",
```

ev-br:
> Perhaps it may be better to prefer calling methods on the tensors rather than torch. free functions to make it a little bit less PyTorch-specific? Or perhaps it doesn't matter? WDYT

So using methods not torch-namespaced functions is more idiomatic in pytorch? 
Here we'll need `torch.as_tensor` all over the place because numpy is very liberal in accepting scalars and other array_likes where torch wants `Tensor`s

autogen/gen_ufuncs.py
```diff
@@ -0,0 +1,127 @@
...
+namespace = np
+
+dct = grab_namespace(namespace)
+
+
+# SKIP these (need special handling)
+skip = {np.frexp, np.modf,    # non-standard unary ufunc signatures
+        np.isnat,
+        np.invert,      # bitwise NOT operator
+        np.spacing,     # niche, does not have a direct equivalent
```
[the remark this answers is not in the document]
lezcano: I don't think we have an equivalent to this one in PyTorch, no, so skipping is fine here.

wrapper.py
```diff
@@ -0,0 +1,1278 @@
...
+pytorch tensors.
+"""
+import numpy as np
+import torch
+
+# Things to decide on (punt for now)
+#
+# 1. Q: What are the return types of wrapper functions: plain torch.Tensors or
+#       wrapper ndarrays.
+#    A: Tensors, apparently
```
ezyang, line 13: You should return torch.numpy.ndarray, how are you going to compose these functions into bigger ones otherwise?

ezyang:
It would be nice for there to be less code (e.g., I hope for cases where APIs match exactly, metaprogramming can make this a single line of code)

wrapper.py
```diff
@@ -0,0 +1,1278 @@
...
+        raise NotImplementedError
+    result = torch.asarray(object, dtype=dtype, copy=copy)
+    ndim_extra = ndmin - result.ndim
+    if ndim_extra > 0:
+        result = result.reshape((1,)*ndim_extra + result.shape)
+    return result
+
+
+def copy(a, order='K', subok=False):
+    if order != 'K' or subik:
```
ezyang, line 60: Subok

wrapper.py
```diff
@@ -0,0 +1,1278 @@
...
+
+def full(shape, fill_value, dtype=None, order='C', *, like=None):
+    if order != 'C' or like is not None:
+        raise NotImplementedError
+    return torch.full(shape, fill_value, dtype=dtype)
+
+
+def full_like(a, fill_value, dtype=None, order='K', subok=True, shape=None):
+    if order != 'K' or not subok:
+        raise NotImplementedError
```
ezyang, line 99: I can tell this order stuff is going to be a pain already

wrapper.py
```diff
@@ -0,0 +1,1278 @@
...
+
+# Things to decide on (punt for now)
+#
+# 1. Q: What are the return types of wrapper functions: plain torch.Tensors or
+#       wrapper ndarrays.
+#    A: Tensors, apparently
+#
+# 2. Q: Default dtypes: numpy defaults to float64, pytorch defaults to float32
+#    A: Stick to pytorch defaults?
+#    NB: numpy recommends `dtype=float`?
```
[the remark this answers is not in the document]
rgommers: `jax.numpy` doesn't even build with `float64` dtype support by default. It has an API matching `numpy`, but the default dtype is `float32`. To get 64-bit floats (from https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#double-64bit-precision):
```python
from jax.config import config
config.update("jax_enable_x64", True)
x = random.uniform(random.PRNGKey(0), (1000,), dtype=jnp.float64)  # note: explicitly ask for float64
x.dtype # --> dtype('float64')
```

wrapper.py
rgommers: +1, I mentioned the same in gh-1.

wrapper.py
rgommers: That will stay AFAIK. `unpack_scalar` to make this easier to understand seems like a good idea to me.

torch_np/_wrapper.py
```diff
@@ -0,0 +1,788 @@
...
+        raise NotImplementedError 
+    if ddof == 0:
+        unbiased = False
+    elif ddof == 1:
+        unbiased = True
+    else:
+        raise ValueError(f"std: ddof = {ddof}")
+    if not torch.is_floating_point:
+        a = a * 1.0
+    return torch.std(a, axis, keepdim=bool(keepdims), out=out)
```
peterbell10, line 678:
```suggestion
    if not torch.is_floating_point(a):
        a = a * 1.0
    return torch.std(a, axis, correction=ddof, keepdim=bool(keepdims), out=out)
```

torch_np/_wrapper.py
```diff
@@ -0,0 +1,788 @@
...
+        raise NotImplementedError 
+    if ddof == 0:
+        unbiased = False
+    elif ddof == 1:
+        unbiased = True
+    else:
+        raise ValueError(f"std: ddof = {ddof}")
+    if not torch.is_floating_point:
+        a = a * 1.0
+    return torch.var(a, axis, keepdim=bool(keepdims), out=out)
```
peterbell10, line 695:
```suggestion
    if not torch.is_floating_point(a):
        a = a * 1.0
    return torch.var(a, axis, correction=ddof, keepdim=bool(keepdims), out=out)
```

ev-br: Thanks Peter!

torch_np/_funcs_impl.py
```diff
@@ -0,0 +1,1924 @@
...
+    t_b = _util.cast_if_needed(t_b, dtype)
+
+    result = torch.vdot(t_a, t_b)
+
+    if is_half:
+        result = result.to(torch.float16)
+    elif is_bool:
+        result = result.to(torch.bool)
+
+    return result.item()
```
lezcano, line 1175: This should be just `result` as it'll return a 0-dim tensor (like reductions).

torch_np/_funcs_impl.py
```diff
@@ -0,0 +1,1924 @@
...
+        # For more than d=2, the strided formula is only valid for arrays with
+        # all dimensions equal, so we check first.
+        s = a.shape
+        if s[1:] != s[:-1]:
+            raise ValueError("All dimensions of input must be of equal length")
+        sz = torch.as_tensor(a.shape[:-1])
+        step = 1 + (torch.cumprod(sz, 0)).sum()
+
+    # Write the value out into the diagonal.
+    a.ravel()[:end:step] = val
```
lezcano, line 1142: Why not
```suggestion
    idx = diag_indices_from(a)
    a[idx] = val
```

lezcano (commented):
I was going to submit fixes for these, but I hit https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/111. I'll try to set up my dev env another time and send fixes for a few of these things I find.

torch_np/_ndarray.py
```diff
@@ -0,0 +1,520 @@
...
+
+    # lists of ndarrays: [1, [2, 3], ndarray(4)] convert to lists of lists
+    if isinstance(obj, (list, tuple)):
+        a1 = []
+        for elem in obj:
+            if isinstance(elem, ndarray):
+                a1.append(elem.tensor.tolist())
+            else:
+                a1.append(elem)
+        obj = a1
```
lezcano, line 473: Couldn't we have a list of lists of arrays? In that case this would not work. If we want that to work, this bit needs to be recursive.

torch_np/_ndarray.py
```diff
@@ -0,0 +1,520 @@
...
+    def __rfloordiv__(self, other):
+        return _ufuncs.floor_divide(other, self)
+
+    def __ifloordiv__(self, other):
+        return _ufuncs.floor_divide(self, other, out=self)
+
+    __divmod__ = _ufuncs.divmod
+
+    # power, self**exponent
+    __pow__ = __rpow__ = _ufuncs.float_power
```
lezcano, line 296:
```suggestion
    __pow__ = _ufuncs.float_power
```

torch_np/_ndarray.py
```diff
@@ -0,0 +1,520 @@
...
+        return _ufuncs.bitwise_or(self, other, out=self)
+
+    # xor, self ^ other
+    __xor__ = __rxor__ = _ufuncs.bitwise_xor
+
+    def __ixor__(self, other):
+        return _ufuncs.bitwise_xor(self, other, out=self)
+
+    # bit shifts
+    __lshift__ = __rlshift__ = _ufuncs.left_shift
```
lezcano, line 330: bit shifts (left and right) are not commutative, so the r-looking variants should be made explicit as well.

torch_np/_ndarray.py
```diff
@@ -0,0 +1,520 @@
...
+        return _funcs.reshape(self, shape, order=order)
+
+    ravel = _funcs.ravel
+    flatten = _funcs._flatten
+
+    def resize(self, *new_shape, refcheck=False):
+        # ndarray.resize works in-place (may cause a reallocation though)
+        self.tensor = _funcs_impl._ndarray_resize(
+            self.tensor, new_shape, refcheck=refcheck
+        )
```
lezcano, line 377: Should we move the implementation of this one to here?

PR #3, opened 2022-12-13
WIP: use wrapper arrays at interfaces

ev-br:
closes gh-1


wrapper.py
```diff
@@ -37,15 +37,33 @@
...
-    _util.subok_no_ok(like)
+    _util.subok_not_ok(like)
     if order is not None:
         raise NotImplementedError
-    return torch.asarray(a, dtype=dtype)
+
+    # This and array(...) are the only places which talk to ndarray directly.
+    # The rest goes through asarray (preferred) or array.
+    out = ndarray()
+    out._tensor = torch.asarray(a, dtype=dtype)
```
ev-br, line 65: @lezcano could you enlighten me on the difference between torch.asarray and torch.as_tensor?

lezcano: They are equivalent. cf.
https://github.com/pytorch/pytorch/blob/e54c6c2870d6f1e318cdd86fa94c37ad22bff22a/test/test_tensor_creation_ops.py#L3908-L3911

ev-br: Thanks! Here's one puzzler though:

```
In [43]: x = np.float64(0.5)

In [44]: torch.asarray(x)
<ipython-input-44-62b3e1ccf04c>:1: UserWarning: The given buffer is not writable, and PyTorch does not support non-writable tensors. This means you can write to the underlying (supposedly non-writable) buffer using the tensor. You may want to copy the buffer to protect its data or make it writable before converting it to a tensor. This type of warning will be suppressed for the rest of this program. (Triggered internally at /home/conda/feedstock_root/build_artifacts/pytorch-recipe_1670076225403/work/torch/csrc/utils/tensor_new.cpp:1563.)
  torch.asarray(x)
Out[44]: tensor([0.0000, 1.7500])

In [45]: torch.as_tensor(x)
Out[45]: tensor(0.5000, dtype=torch.float64)
```

Is is supposed to handle numpy scalars?

_unary_ufuncs.py
```diff
@@ -4,9 +4,10 @@
 import torch
 
 import _util
+from _ndarray import asarray_replacer_1
 
 
-
+@asarray_replacer_1
 def absolute(x, /, out=None, *, where=True, casting='same_kind', order='K',
```
lezcano, line 11: Note that in PyTorch we still support Python 3.7, so perhaps it'd be better not to use positional only args? cc @ezyang.

_ndarray.py
```diff
@@ -0,0 +1,103 @@
...
+
+def asarray_replacer_1(func):
+    """Asarray the *first* arg, so that the wrapped `func` operates on a
+       torch.Tensor. Then asarray the result.
+    """
+    @functools.wraps(func)
+    def wrapped(x, *args, **kwds):
+        x_tensor = asarray(x).get()
+        return asarray(func(x_tensor, *args, **kwds))
+    return wrapped
```
lezcano, line 102: Perhaps we want to wrap any `np.ndarray` that's given to us in the signature? This way this decorator could be written generically for all the functions.

lezcano: If the obs above makes sense, then we could potentially apply the wrapper generically to all the functions defined in the module at the very end of the module, that way we make sure we don't miss any of them and we'd save quite a few lines.

wrapper.py
lezcano: No idea about that edge case, but if  `as_tensor` works, I'd simply use `torch.as_tensor`.
cc @ysiraichi who wrote the test I linked above.

_ndarray.py
ev-br: Great idea, thanks! I'd prefer to defer it a bit though, to first work through the numpy API patterns (it's... uhm... organically grown).

_unary_ufuncs.py
lezcano: Note that we want to be NumPy compatible. As such, implementing a superset of NumPy's functionality would also do. Dropping positional-only args would achieve this. But again, let's wait to see what Ed thinks, perhaps PyTorch is planning to drop Python 3.7 support, idk.

_ndarray.py
rgommers: Is decorating a function that returns a Tensor with this replacer actually shorter than not doing so? I'm not yet seeing that yet - only if it has to do more work than wrapping the output arrays. This seems equally short and clearer:
```python
class _ndarray:
    # class name here has to have an underscore probably, so the public ndarray object can use this
    # one but have a signature for __init__ that matches numpy (this is a small detail only).
    def __init__(self, tensor):
        self._tensor = tensor


def absolute(...):
    return _ndarray(result)
```

rgommers: On the other hand, this code block is repeated a lot and seems to need factoring out to be generic across ufuncs:
```python
    _util.subok_not_ok(subok=subok)
    if order != 'K' or casting != 'same_kind' or not where:
        raise NotImplementedError
    if out is not None:
      # XXX dtypes, casting
        out = out.to(dtype)
```

lezcano: Note that the decorator aso wraps the first argument (and potentially others). Also, as discussed above, the decorator could potentially be applied to all functions in the module a posteriori so that we don't miss any of them. This would very much reduce the amount of boilerplate needed all across.

wrapper.py
ysiraichi: Not an edge case, but a bug! It should work fine if you specify the dtype, though.

```
>>> torch.asarray(x, dtype=torch.float64)
tensor([0.5000], dtype=torch.float64)
```

_ndarray.py
rgommers: > Note that the decorator aso wraps the first argument (and potentially others).

It's not clear to me that it has to do that, or that it's better. The current implementation also explicitly does `torch.as_tensor(x)` internally anyway (maybe a leftover).

The problem with these kinds of decorators, unless they do more than wrap inputs and outputs, is that they're hard to read for no real gain. Right now, `out` isn't handled for example - and that can also be an array. There will be lots of one-offs where some other argument can be an array, and there's no decorator for exactly that thing. So now you have to go write yet another decorator with a slight twist, or mix decorator with manual wrapping.




> This would very much reduce the amount of boilerplate needed all across.

I think it makes sense to figure out if this is actually true. Here is how I'd write `absolute`, and I like how it looks better than with the decorator approach:
```python

def handle_ufunc_kwargs(subok=None, order=None, out=None, dtype=None):
    # This is mostly factoring out what was already written, it's not complete
    _util.subok_not_ok(subok=subok)
    if order != 'K' or casting != 'same_kind' or not where:
        raise NotImplementedError

    if out is not None:
        out = as_tensor(out)  # unwrap
        if dtype is not None:
            if dtype != out.dtype:
                raise ValueError("given `out` array doesn't have the right dtype")

    return out


def absolute(x, /, out=None, *, where=True, casting='same_kind', order='K',
          dtype=None, subok=False, **kwds):
    x = as_tensor(x)
    out = handle_ufunc_kwargs(subok=subok, order=order, out=out, dtype=dtype)

    result = torch.absolute(x, out=out)
    if dtype is not None:
        result = result.to(dtype)
    
    return _ndarray(result)
```

_ndarray.py
```diff
@@ -0,0 +1,144 @@
...
+    def __str__(self):
+        return str(self._tensor).replace("tensor", "array_w")
+
+    __repr__ = __str__
+
+
+    ### arithmetic ###
+
+    def __add__(self, other):
+        return self._tensor__add__(other)
```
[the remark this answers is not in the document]
ev-br: Thanks! This part is incomplete, untested yet.

_unary_ufuncs.py
rgommers: > This is then a question of what numpy version we want to track here.

1.24 is coming out in 1-2 weeks, so probably that. Won't be much different than 1.23.x though. Although tracking evolving function signatures is always a pain, and we'll need some way of handling that at some point.

wrapper.py
ysiraichi: Correction: it's not a bug... I will look into this.

```
In [6]: torch.asarray(np.array([1, 2, 3], dtype=np.float64))
Out[6]: tensor([1, 2, 3])
```

_ndarray.py
ev-br: The ufunc code block is actually autogenerated from template, https://github.com/Quansight-Labs/numpy_pytorch_interop/blob/ndarrays_at_interface/autogen/gen_ufuncs.py#L72 
and it's of course a stub waiting to handle dtypes and out arrays (shapes, types etc), also casting. 

Given how this develops for now, I'd like to push the current approach for a bit more to see various patterns in numpy. E.g., I'd rather not detail the ufunc handler yet, because e.g. `subok` is used in both ufuncs and non-ufuncs. Same for generalizing the decorators or other approaches, let's first see what is needed, then clean up and consolidate. I am not yet certain that what is here now is truly representative, need to convert some more functions.

Whether a single decorator can be applied to all args, I suspect not because array_like treatment differs by function:
- func(array_like, array_like) -- e.g. binary ufuncs
- func(array_like, tuple) -- e.g. `np.reshape(array, new_shape)`
- func(*array_likes) -- e.g. `atleast_1d` 

One other thing I'd rather keep rolling with is not using `ndarray` class (with or with or without a leading underscore) for wrapping the outputs. The public `ndarray` class will need to have numpy-compatible signature. So for now I'm using `asarray` as a central thing which knows about the internals about the `ndarray` class. Let me see how far this takes us.

rgommers: > One other thing I'd rather keep rolling with is not using `ndarray` class (with or with or without a leading underscore) for wrapping the outputs.

That's fine to defer I'd say, it's easy to tweak later and it's a ~10 characters change.

wrapper.py
ysiraichi: Long story short: `torch.asarray` works with Numpy scalars, but it doesn't infer the type. Thus, if you don't specify the dtype, it will assume the default dtype.

**Explanation:** since Numpy scalars are not Numpy arrays, they end up being treated as Python objects that implement the protocol buffer ([as documented](https://pytorch.org/docs/stable/generated/torch.asarray.html)). i.e. it will rely on the given `dtype` argument (if given) and interpret the data as such.

**Discussion:** I guess the whole point of `asarray` is to try not to copy. I don't think it makes much sense using it for scalars... That said, it's an non-intuitive result that has an easy fix.

@lezcano @rgommers 
What do you think?

lezcano: I reckon that extending `torch.asarray` to also handle numpy scalars as expected wouldn't hurt, but let's see what Ralf has to say.

_ndarray.py
ev-br: > x = as_tensor(x)

BTW, this does not handle ndarrays.

_unary_ufuncs.py
ezyang: we should be dropping 3.7 soon

ezyang: ok clarification: we will drop it when colab drops it lol

ev-br:
Okay, as a PR to close gh-1 I think this PR has run its course. We clearly have all public functions consume array_likes and return wrapper ndarrays, so what is currently in main is obsolete. Implementation details are still in flux, and will stay in flux for a while. Will revisit them when after the next milestone, when the API coverage is representative.

PR #6, opened 2022-12-18
WIP: bare-bones dtypes; can_cast and result_type

ev-br:
Make wrappers for the intersection of torch-supported and numpy-supported dtypes. This is very rudimentary, no attempt is made to follow the actual numpy dtype hierarchy (DType, _DTypeMeta etc). Also this does not really distinguish between array scalars and dtypes (classes or instances). Not sure how deep into this rabbit hole we need to fall, really.

For dtype casting, 

1. No value-based casting (this is a feature, not bug!)
2. Ignoring the difference between scalars and arrays for dtype promotion (`3.0*array(..., dtype="float32")` -> dtype("float64") or float32 etc).
3. track numpy 1.23.5 otherwise

TODO : 

- [ ] actually use it for casting inputs and outputs in ufuncs and relatives
- [ ] dtype aliases: np.dtype('i8'), conversion from python types (`dtype=float`), typecodes etc


ev-br:
Superseded by gh-7.

PR #7, opened 2022-12-19
WIP: add bare-bones dtypes, rework ndarray.methods / free functions split

ev-br:
Use np.concatenate tests for dogfooding.


torch_np/_wrapper.py
```diff
@@ -258,13 +268,60 @@ def corrcoef(x, y=None, rowvar=True, bias=NoValue, ddof=NoValue, *, dtype=None):
...
+                # while numpy errors out
+                # TODO: centralize validation of out (incl this numpy/torch discrepancy)
+                result = torch.cat(tensors, axis, out=out_tensor)
+        except UserWarning:
+            raise ValueError("output array has wrong shape or dimensionality")
+    except (IndexError, RuntimeError):
+        raise AxisError
+    if out is not None:
+        return out
+    return asarray(result)
```
ev-br, line 324: This `concatenate` functions is a prototype for a top-level namespace function without a corresponding ndarray method.

torch_np/_dtypes.py
```diff
@@ -0,0 +1,174 @@
...
+# TODO: 1. define torch_np dtypes, make this work without numpy.
+#       2. mimic numpy's various aliases (np.half == np.float16, dtype='i8' etc)
+#       3. convert from python types: np.ones(3, dtype=float) etc
+
+import builtins
+import torch
+
+# Define analogs of numpy dtypes supported by pytorch.
+
+class dtype:
```
ev-br, line 13: the `dtype` class is a super-minimal duck-typing analog of numpy dtypes. Instances are not singletons! `dtype("int64") == dtype("int64")` but these are two different instances.

torch_np/_dtypes.py
```diff
@@ -0,0 +1,174 @@
...
+        raise TypeError
+
+
+
+# The casting below is defined *with dtypes only*, so no value-based casting!
+
+# These two dicts are autogenerated with autogen/gen_dtypes.py,
+# using numpy version 1.23.5.
+
+_can_cast_dict = {
```
ev-br, line 144: Implementation detail: no structured casting between categories, no differentiating between arrays and scalars (scalar can upcast arrays).

torch_np/_ndarray.py
```diff
@@ -29,24 +38,58 @@ def ndim(self):
...
-        return self._tensor.dtype
+        return _dtypes.dtype_from_torch(self._tensor.dtype)
 
     @property
     def strides(self):
         return self._tensor.stride()   # XXX: byte strides
 
+    @property
+    def base(self):
+        return self._base
```
ev-br, line 49: This might create mighty gc cycles,not sure if this is a problem. Or how to avoid it if it is.

lezcano: Why would it create gc cycles? I think that the ownership structure here is clear.

torch_np/_wrapper.py
```diff
@@ -114,14 +121,17 @@ def arange(start, stop=None, step=1, dtype=None, *, like=None):
...
+    return asarray(torch.arange(start, stop, step, dtype=torch_dtype))
 
 
 def empty(shape, dtype=float, order='C', *, like=None):
     _util.subok_not_ok(like)
     if order != 'C':
         raise NotImplementedError
-    return asarray(torch.empty(shape, dtype=dtype))
+    dtype = _dtypes.dtype(dtype)
+    torch_dtype = _dtypes.torch_dtype_from(dtype)
```
lezcano, line 134: We may want to put these two into a function, given that we're going to use this pattern left right and centre.

torch_np/_wrapper.py
```diff
@@ -258,13 +268,60 @@ def corrcoef(x, y=None, rowvar=True, bias=NoValue, ddof=NoValue, *, dtype=None):
...
+    if out is not None:
+        if dtype is not None:
+            # mimic numpy
+            raise TypeError("concatenate() only takes `out` or `dtype` as an "
+                            "argument, but both were provided.")
+        if not isinstance(out, ndarray):
+            raise ValueError("'out' must be an array")
+        out_tensor = out.get()
+    else:
+        out_tensor = out
```
lezcano, line 281: Fantastic candidate to be factored out into a function.

lezcano (commented):
Just leaving here what we already discussed:

It's clear that the current approach does not scale if for a for a function like `np.concatenate` which has a torch equivalent given by `torch.cat` we need to write 50+ highly non-trivial LOC.

I would love to be able to implement this and other functions as:
```python
@common_inputs
def concatenate(ar_tuple, axis=0, out=None, dtype=None, casting="same_kind"):
      if axis is None:
        ar_tuple = tuple(tensor.ravel() for tensor in ar_tuple)
        axis = 0
      return torch.cat(ar_tuple, axis)
```
In `common_inputs` we would, in this case:
- Wrap the arrays in `ar_tuple` into `Tensor`s.
- Cast them as per `casting`
- 
Then, after the function is called, we would handle `out` via `out.copy_(result)` (and a cast if necessary).

The current approach of always setting `out=out_tensor` is not great due to how PyTorch handles dispatching. If you set the `out=` argument, even if it's to pass `out=None`, it will dispatch to the `out=` variant of the operator. Now, these are not differentiable, which is a pain. The `copy_` approach is not only generic, but also differentiable, which is neat. Then TorchInductor (the fuser) will do the heavy lifting of fusing and removing the unnecessary ops.

ev-br:
> The current approach of always setting out=out_tensor is not great due to how PyTorch handles dispatching. If you set the out= argument, even if it's to pass out=None, it will dispatch to the out= variant of the operator. Now, these are not differentiable, which is a pain. The copy_ approach is not only generic, but also differentiable, which is neat. Then TorchInductor (the fuser) will do the heavy lifting of fusing and removing the unnecessary ops.

Thanks, that's great to know! 
Is this what you had in mind then? https://github.com/Quansight-Labs/numpy_pytorch_interop/blob/rework_unary_ufunc/torch_np/_unary_ufuncs.py#L46

(L46 to L61)

torch_np/_wrapper.py
ev-br: Indeed, if only the logic generalizes. E.g. unary ufuncs seem to accept both `dtype=` and `out=`. (Not that I understand what `np.sin(x, dtype=..., out=...)` does, but it definitely accepts both)

rgommers: That's fine, right? If both are given, then `dtype == `out.dtype` should be true, otherwise you raise. And if it's true, there is no issue. I'd expect that to generalize to all functions.

ev-br: Kind of.

```
In [550]: np.add(np.ones(3), np.ones(3), dtype=float, out=np.empty(3))     # ufuncs accept both
Out[550]: array([2., 2., 2.])

In [551]: np.concatenate((np.ones(3), np.ones(4)), dtype=float, out=np.empty(7))    # concatenates does not
TypeError                                 Traceback (most recent call last)
Input In [551], in <cell line: 1>()
----> 1 np.concatenate((np.ones(3), np.ones(4)), dtype=float, out=np.empty(7))

File <__array_function__ internals>:180, in concatenate(*args, **kwargs)

TypeError: concatenate() only takes `out` or `dtype` as an argument, but both were provided.

In [554]: np.abs(np.ones(3), dtype='int32', out=np.empty(3, dtype='complex'))      # what is this doing?
Input In [554], in <cell line: 1>()
----> 1 np.abs(np.ones(3), dtype='int32', out=np.empty(3, dtype='complex'))

UFuncTypeError: Cannot cast ufunc 'absolute' input from dtype('float64') to dtype('int32') with casting rule 'same_kind'

In [555]: np.abs(np.ones(3), dtype='float32', out=np.empty(3, dtype='complex'))    # but not always
Out[555]: array([1.+0.j, 1.+0.j, 1.+0.j])
```

what `dtype=` does for ufuncs is a bit of a mystery to me (how does this interact with `.types` of registered loops), but it definitely accepts dtype != out.dtype. Maybe safe to side-step as long as actual computations are offloaded to pytorch and it does it own casting dance internally. 

So the suggestion for ufuncs is to ignore the `dtype` if `out is not None`. Or maybe ignore it completely even.

rgommers: This looks like a bug, or at least questionable behaviour, in numpy:
```python
>>> x = np.ones(3)  # has dtype float64
>>> np.add(x, x, dtype=np.float64, out=np.empty(3, dtype=np.float16))
array([2., 2., 2.], dtype=float16)
```
I can't find an existing open issue (but it could exist, this is hard to search for), nor a note in the docs on this. I think what is happening is that the semantics of `dtype` is not what you'd expect. Based on common usage experience and common sense, you'd want it to be the _output dtype_. It usually is, but that's only a consequence of its actual semantics: it selects the _dtype of the inner loops_ of the implementation underlying the function. 

For non-ufuncs, the `dtype` keyword is typically documented as determining the output dtype. And for ufuncs, that's effectively how it's used. The "assume users know numpy internals, and let them select inner loop casts" makes very little sense.

> So the suggestion for ufuncs is to ignore the `dtype` if `out is not None`. Or maybe ignore it completely even.

I think this is exactly an example like @lezcano suggested. Whatever we do, let's use a consistent design rather than follow the quirks of numpy exactly. Imho best to not have any distinction between what ufuncs and non-ufunc functions do here.

ev-br: I can guess a use case for the accumulator dtype being wider than the input dtype, as in `np.add(x, x, dtype=float64)` with `x.dtype==np.float16`. If we were designing it from scrach, I'd go for overloading the `out` argument into either an array to place the result into, or a dtype for a fresh array. But since we are not :-), and given that pytorch does not support `torch.add(x, x, dtype=....)` anyway, the two options I see are either raise or ignore. Raising seems less numpy-compatible, can make porting from numpy harder.

ev-br: IOW, for numpy itself, `dtype=...` might as well belong to "other obscure keyword args", somewhere together with `extobj`, https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs. Did somebody start planning numpy 2.0? :-).

ev-br:
Okay, making some progress. This does yet look like @lezcano 's desired result, but it's getting closer.

torch_np/_wrapper.py
```diff
@@ -258,13 +268,60 @@ def corrcoef(x, y=None, rowvar=True, bias=NoValue, ddof=NoValue, *, dtype=None):
...
+            with warnings.catch_warnings():
+                warnings.simplefilter('error', UserWarning)
+                # torch emits a UserWarning if the out tensor has wrong size,
+                # while numpy errors out
+                # TODO: centralize validation of out (incl this numpy/torch discrepancy)
+                result = torch.cat(tensors, axis, out=out_tensor)
+        except UserWarning:
+            raise ValueError("output array has wrong shape or dimensionality")
+    except (IndexError, RuntimeError):
+        raise AxisError
```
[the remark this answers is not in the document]
ev-br: This is gone, happy to see the back of it, too.

torch_np/_wrapper.py
rgommers: Yes, the accumulator thing is probably a/the reason for why it was added way back when. But that's extremely niche, and in the few cases where you actually need it, it should be done by default. 

> Did somebody start planning numpy 2.0? :-).

Yes, actually - a NEP should arrive shortly. 1.26 -> 2.0 (Dec'23)

lezcano: In PyTorch we have a `dtype=` kwarg for reductions. This makes complete sense. Now, the benefits of having it for a non-reduction operator I think that are fairly niche and better dropped out.

You can have a look at the semantics of `dtype=` (and of `out=` and other kwargs) within the context of a reduction in the generic implementation of a reduction in PrimTorch:
https://github.com/pytorch/pytorch/blob/fd3a7264ae80332ba5ec8f60446e6d7a2c2276c1/torch/_refs/__init__.py#L1992-L2003

lezcano:
>Thanks, that's great to know!
Is this what you had in mind then? https://github.com/Quansight-Labs/numpy_pytorch_interop/blob/rework_unary_ufunc/torch_np/_unary_ufuncs.py#L46
(L46 to L61)

It's getting closer, but I think we should be able to do better! I'd love to see an implementation of `absolute` that reads (something like):
```python
@common_args(args=("out",), kwargs=("where", "casting", "order", "dtype", "subok"))
def absolute(x):
    return torch.absolute(x)
```
Then `args` and `kwargs` would have a list of strings they understand. For those they don't understand they'd error out. This list can be an enum, whatever works.
This `common_args` would implement most of the shared logic of all these functions. For example, it will wrap any `ndarray` input/output.

Then, each kwarg should be implementable as an independent decorator (as in, takes a function and wraps it into some extra logic before and after it). An example of such decorator is `out_wrapper` in PrimTorch.
https://github.com/pytorch/pytorch/blob/fd3a7264ae80332ba5ec8f60446e6d7a2c2276c1/torch/_prims_common/wrappers.py#L176
Note that most of the complexity in that decorator stems from having to put the correct type annotations in the function. The logic of the decorator itself is rather short and concise. In general, that decorator implements the canonical semantics for `out=` in PyTorch.

Please tell me if this approach makes sense and it's clear :)

ev-br:
Let's keep the `@common_args` decorator as a north star to follow and keep iterating until we have a reasonable test coverage. With that, we'll be able to clean up, consolidate and refactor various numpy patterns.

PR #8, opened 2022-12-26
Implement reduction operations

ev-br:
- [x] count_nonzero
- [x] any, all
- [x] argmin, argmax
- [x] min, max (or amin, amax --- no indices)
- [ ] var, std, median
- [x] mean, sum

The plan is to have a minimal set, then deduplicate.


torch_np/_ndarray.py
```diff
@@ -147,11 +223,105 @@ def ravel(self, order='C'):
...
+    @axis_out_keepdims_wrapper
+    def min(self, axis=None, out=None, keepdims=NoValue, initial=NoValue,
+             where=NoValue):
+        if where is not None:
+            raise NotImplementedError
+        if initial is not None:
+            raise NotImplementedError
+
+        result = self._tensor.amin(axis)
+        return result
```
ev-br, line 278: @lezcano could you please take a look when you have time? Two points I'm not sure about interactions with pytorch:

- it's convenient to use `torch.amax` here, not `torch.max`, because `amax` supports tuple axes while `max` doesn't; there is however a cryptic comment in https://pytorch.org/docs/stable/generated/torch.amax.html?highlight=amax#torch.amax, _amax/amin evenly distributes gradient between equal values, while max(dim)/min(dim) propagates gradient only to a single index in the source tensor._ --- is this relevant to use here?

- since `dim / keepdim` handling is not entirely uniform across reductions, this PR opts to not propagating `keepdim` to torch at all, and handle it in the `@axis_out_keepdims_wrapper`. Does it play well with dynamo or inductor or who is the consumer of this?

Other than these, I think this PR is a reasonable approximation to generic handling of reductions, so a review would be appreciated if you have time.

torch_np/_wrapper.py
```diff
@@ -179,14 +177,28 @@ def logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None, axis=0):
...
+    if step == 0:
+        raise ZeroDivisionError
+    if stop is None and start is None:
+        raise TypeError
     if stop is None:
-        # arange(stop)
-        stop = start
+        # XXX: this breaks if start is passed as a kwarg:
+        # arange(start=4) should raise (no stop) but doesn't
+        start, stop = 0, start
```
ev-br, line 189: It's a funny little glitch how both pytorch and numpy `arange` signatures cannot be easily represented in pure python without going fully manual on `arange(*args, **kwargs)`. For instance, https://pytorch.org/docs/stable/generated/torch.arange.html lists `torch.arange(start=0, end, step=1, *, ...)` which is not valid python (`end` does not have a default). While this breaks at least one numpy test, I'd rather leave it as is :-).

torch_np/_ndarray.py
lezcano: Always prefer `amax` over `max` because of that point about gradients. This will be relevant once we start considering using this wrapper to compute derivatives of NumPy programs. For reference, the comment references the case when you compute the maximum of a tensor with repeated values. The function `max` is not differentiable at these points, but it's convex, so we compute its subgradient of max norm.

About `dim`/`keepdim`, do you mean that it's not uniform on PyTorch or on NumPy? If it's the former, can you give an example? We should fix it in PyTorch if that's the case.

torch_np/_wrapper.py
lezcano: Your approach goes in the right direction, but we may be able to do better?
https://github.com/pytorch/pytorch/blob/2b52db9c953d063db7b46c12f4df35b47aca4381/torch/_refs/__init__.py#L4221-L4237

torch_np/_ndarray.py
ev-br: Thanks, will keep using `amin, amax` then. 

dim/keepdim comment is about pytorch:

```
In [33]: torch.any([0, 1, 2, 3], keepdim=True)
TypeError                                 Traceback (most recent call last)
Input In [33], in <cell line: 1>()
----> 1 torch.any([0, 1, 2, 3], keepdim=True)

TypeError: any() received an invalid combination of arguments - got (list, keepdim=bool), but expected one of:
 * (Tensor input, *, Tensor out)
 * (Tensor input, int dim, bool keepdim, *, Tensor out)
 * (Tensor input, name dim, bool keepdim, *, Tensor out)
```

but e.g. `amax` accepts keepdim without an explicit axis (but refuses to convert a list into a tensor):

```
In [34]: torch.amax([0, 1, 2, 3], keepdim=True)                    # oops, lists are not accepted
TypeError                                 Traceback (most recent call last)
Input In [34], in <cell line: 1>()
----> 1 torch.amax([0, 1, 2, 3], keepdim=True)

TypeError: amax(): argument 'input' (position 1) must be Tensor, not list

In [35]: torch.amax(torch.as_tensor([0, 1, 2, 3]), keepdim=True)
Out[35]: tensor([3])
```

torch_np/_wrapper.py
ev-br: Applying this patch

```
(pytorch) br@gonzales:~/sweethome/proj/scipy/torch_np_compat$ git diff
diff --git a/torch_np/_wrapper.py b/torch_np/_wrapper.py
index bea4c2f..82444d5 100644
--- a/torch_np/_wrapper.py
+++ b/torch_np/_wrapper.py
@@ -177,18 +177,20 @@ def logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None, axis=0):
     return asarray(torch.logspace(start, stop, num, base=base, dtype=dtype))
 
 
-def arange(start=None, stop=None, step=1, dtype=None, *, like=None):
+def arange(start=0, stop=None, step=1, dtype=None, *, like=None):
     _util.subok_not_ok(like)
     if step == 0:
         raise ZeroDivisionError
-    if stop is None and start is None:
-        raise TypeError
-    if stop is None:
-        # XXX: this breaks if start is passed as a kwarg:
+#    if stop is None and start is None:
+#        raise TypeError
+#    if stop is None:
+#        # XXX: this breaks if start is passed as a kwarg:
         # arange(start=4) should raise (no stop) but doesn't
+#        start, stop = 0, start
+#    if start is None:
+#        start = 0
+    if stop is None:
         start, stop = 0, start
-    if start is None:
-        start = 0
``` 

results in

```
$ pytest torch_np/tests/test_function_base.py -v -x
====================================== test session starts =======================================
platform linux -- Python 3.9.0, pytest-7.2.0, pluggy-1.0.0 -- /home/br/miniconda3/envs/pytorch/bin/python3.9
cachedir: .pytest_cache
rootdir: /home/br/sweethome/proj/scipy/torch_np_compat
plugins: cov-4.0.0
collected 12 items                                                                               

torch_np/tests/test_function_base.py::TestCopy::test_basic PASSED                          [  8%]
torch_np/tests/test_function_base.py::TestCopy::test_order SKIPPED (ndarray.flags not ...) [ 16%]
torch_np/tests/test_function_base.py::TestArange::test_infinite PASSED                     [ 25%]
torch_np/tests/test_function_base.py::TestArange::test_nan_step PASSED                     [ 33%]
torch_np/tests/test_function_base.py::TestArange::test_zero_step PASSED                    [ 41%]
torch_np/tests/test_function_base.py::TestArange::test_require_range FAILED                [ 50%]

============================================ FAILURES ============================================
_________________________________ TestArange.test_require_range __________________________________

self = <torch_np.tests.test_function_base.TestArange object at 0x7f4655d33580>

    def test_require_range(self):
>       assert_raises(TypeError, np.arange)
E       Failed: DID NOT RAISE <class 'TypeError'>

torch_np/tests/test_function_base.py:60: Failed
```

where the test is straight copy-paste from numpy.

torch_np/_ndarray.py
lezcano: Functions in PyTorch do not accept list in place of tensors, so the second part is expected.

About the first part, yep, `any` is wrong. Now, I don't know if we can fix this in a BC way. WDYT @ezyang @albanD? If it's alright to fix this, I can submit a fix.

For reference, this is the file that binds the C++ API to the Python API. 
https://github.com/pytorch/pytorch/blob/2b52db9c953d063db7b46c12f4df35b47aca4381/aten/src/ATen/native/native_functions.yaml#L9401
In particular, if you look up `amax` and (all the appearances of) `any`, it should be clear what's the issue. For how to read this file, see the `README.md` in that same folder.

torch_np/_wrapper.py
lezcano: Another way of doing this is via overloads. you can see in `native_functions.yaml` the overloads that we have in C++ for this function. It was done like this before in primtorch, but I "fixed it" (clearly not perfectly, as you've found) at some point. The previous implementation went something like
https://github.com/pytorch/pytorch/blob/2f95b61ceaab65580d8dece821693d8b31ecfca8/torch/_refs/__init__.py#L3113-L3138

lezcano:
btw, if you find inconsistencies in PyTorch, feel free to open issues in the PyTorch repo and tag me in them.

torch_np/_wrapper.py
ev-br: It's pretty clear that these are C++ overloads, but this and `dim=..., keepdims=...`. And that they do not cleanly map onto the cartesian product behavior of python's arguments with defaults, unless without a borderline unreasonable amount of effort anyway.

With arange specifically, I'm with you FWIW. Doing `if stop is None: start, stop = 0, start` is close to the only reasonable thing to do. 
When I saw `arange(start=5)` in numpy tests, my first reaction was along the lines of _what in the name of all things graceful is this_.

Anyway, I don't believe this is worth getting to the bottom of, as long as we have better things to do. 

The main take-home message for myself is that positional-only args in python are a thing, since `arange(start, /, stop=None, step=1)` would sidestep this all completely.

lezcano: also, note that, as discussed, implementing a superset of NumPy's functionality is also alright, so in this case we should be good :)

torch_np/tests/test_reductions.py
```diff
@@ -0,0 +1,617 @@
...
+        self.allowed_axes =  [0, 1, 2, -1, -2,
+                              (0, 1), (1, 0), (0, 1, 2), (1, -1, 0)]
+
+
+class TestArgminGeneric(_GenericReductionsTestMixin, _GenericHasOutTestMixin):
+    def setup_method(self):
+        self.func = np.argmin
+        self.allowed_axes =  [0, 1, 2, -1, -2, ]
+
+    @pytest.mark.xfail(reason='XXX: argmin does not allow axis=tuple')
```
lezcano, line 567: Why are these tests xfailed? I don't quite understand the reason.

autogen/numpy_api_dump.py
```diff
@@ -277,9 +256,6 @@ def find_common_type(array_types, scalar_types):
 def fix(x, out=None):
     raise NotImplementedError
```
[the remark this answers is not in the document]
lezcano: Even more, you may want to consider creating a type system and using it here. In PrimTorch it was really helpful because of the linter and to catch errors early, but also when doing metaprogramming.

torch_np/_ndarray.py
```diff
@@ -95,19 +133,38 @@ def __float__(self):
...
 
     def __add__(self, other):
-        return self._tensor.__add__(other)
+        other_tensor = asarray(other).get()
+        return asarray(self._tensor.__add__(other_tensor))
 
     def __iadd__(self, other):
-        return self._tensor.__add__(other)
+        other_tensor = asarray(other).get()
+        return asarray(self._tensor.__add__(other_tensor))
```
lezcano, line 148:
```suggestion
        return asarray(self._tensor.__iadd__(other_tensor))
```

lezcano: Probably it'd be best to implement all these binary dunder methods generically to avoid these slips. There's a similar slip in `__rmul__`.

There's also many others in PyTorch, like `__radd__` `__rdiv__` and so on. I reckon all these also exist in NumPy.

torch_np/_ndarray.py
```diff
@@ -3,11 +3,40 @@
...
+    This decorator implements the generic handling of axis, out and keepdims
+    arguments for reduction functions.
+    """
+    # XXX: move this out of _ndarray.py (circular imports)
+    @functools.wraps(func)
+    def wrapped(a, axis=None, out=None, keepdims=NoValue, *args, **kwds):
+        arr = asarray(a)
+        axis = _helpers.standardize_axis_arg(axis, arr.ndim)
+
+        arr1 = arr
```
lezcano, line 25: why?

torch_np/_util.py
```diff
@@ -82,3 +82,22 @@ def normalize_axis_tuple(axis, ndim, argname=None, allow_duplicate=False):
         else:
             raise ValueError('repeated axis')
     return axis
+
+
+def expand_shape(arr, axis):
```
lezcano, line 87: you probably just want to pass `shape` and do `out_ndim = len(axis) + len(shape)`.

torch_np/_util.py
```diff
@@ -82,3 +82,22 @@ def normalize_axis_tuple(axis, ndim, argname=None, allow_duplicate=False):
...
+    shape_it = iter(arr.shape)
+    shape = [1 if ax in axis else next(shape_it) for ax in range(out_ndim)]
+    return shape
+
+def handle_empty_axis(arr, func, *args, **kwds):
+    """Reductions: func(arr, axis=(), *args, **kwds)
+    """
+    shape = expand_shape(arr, axis=0)
+    arr1 = arr.reshape(shape)
+    return func(arr1, axis=0, *args, **kwds)
```
lezcano, line 102: I think this is not used anymore. It's now inlined in the wrapper that handles those kwargs.

torch_np/_ndarray.py
```diff
@@ -147,11 +223,105 @@ def ravel(self, order='C'):
...
+    @axis_out_keepdims_wrapper
+    def mean(self, axis=None, dtype=None, out=None, keepdims=NoValue, *, where=NoValue):
+        if where is not None:
+            raise NotImplementedError
+
+        if dtype is None:
+            dtype = self.dtype
+        if not _dtypes.is_floating(dtype):
+            dtype = _dtypes.default_float_type()
+        torch_dtype = _dtypes.torch_dtype_from(dtype)
```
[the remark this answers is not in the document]
lezcano: I left a comment. That's not an issue with your implementation but with PyTorch. Thank you for finding this!

torch_np/_util.py
lezcano: Also, if this function is always followed by a reshape, you probably just want to call it `expand` and perform the `view` (`reshape` may copy in PyTorch`) there.
See also https://pytorch.org/docs/stable/generated/torch.Tensor.expand.html

torch_np/_wrapper.py
```diff
@@ -497,10 +492,35 @@ def ravel_multi_index(multi_index, dims, mode='raise', order='C'):
...
+    arr = asarray(a)
+    return nonzero(arr.ravel())[0]
+
+
+def argwhere(a):
+    arr = asarray(a)
+    if arr.ndim == 0:
+        a = atleast_1d(a)
+        # then remove the added dimension
+        return argwhere(a)[:, :0]
```
lezcano, line 510: This cannot be right. Did you mean
```suggestion
        return argwhere(a)[:, 0]
```
?

torch_np/_wrapper.py
```diff
@@ -497,10 +492,35 @@ def ravel_multi_index(multi_index, dims, mode='raise', order='C'):
...
+    return transpose(nonzero(a))
```
lezcano, line 511: The strides of this won't match those of NumPy for sure, but that's alright, let's not care about that for now. Now, it would be good to leave a note in the list of design decisions.

torch_np/_wrapper.py
```diff
@@ -497,10 +492,35 @@ def ravel_multi_index(multi_index, dims, mode='raise', order='C'):
...
+
+
+from ._ndarray import axis_out_keepdims_wrapper
+
+@axis_out_keepdims_wrapper
+def count_nonzero(a, axis=None, *, keepdims=False):
+    # XXX: this all should probably be generalized to a sum(a != 0, dtype=bool)
```
lezcano, line 518: SGTM

torch_np/_wrapper.py
```diff
@@ -570,77 +590,56 @@ def tri(N, M=None, k=0, dtype=float, *, like=None):
...
 def argmin(a, axis=None, out=None, *, keepdims=NoValue):
-    if axis is None:
-        result = torch.argmin(a, keepdim=bool(keepdims))
-    else:
-        result = torch.argmin(a, axis, keepdim=bool(keepdims))
-    if out is not None:
-        out.copy_(result)
-    return result
+    arr = asarray(a)
+    return arr.argmin(axis=axis, out=out, keepdims=keepdims)
```
lezcano, line 600: Should we implement all these in a generic way?

lezcano (commented):
Sorry for the order of the comments, I jumped around quite a lot when doing this review, as the PR was reasonably large.

A general proposal. What about implementing
```python
def reduction(func, *, has_dtype, has_initial, allow_single_axis)
```
and then doing
```python
any = reduction(torch.any, has_dtype=False, has_initial=False, allow_single_axis=True)
all = reduction(torch.all, has_dtype=False, has_initial=False, allow_single_axis=True)
min = reduction(torch.amin, has_dtype=False, has_initial=True, allow_single_axis=False)
...
```
If you want to be a bit less strict, you can have a look at how reductions are implemented in PrimTorch.

torch_np/_ndarray.py
ezyang: Looks like a pretty easy, BC fix. We'd just add another overload without an `int dim` argument.

I wouldn't necessarily block work here on it though; keepdim without a dim argument is easy enough to simulate, no?

ev-br: Exactly. This whole thread started with me asking if it's OK from torchinductor perspective to emulate keepdim in this layer (https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/8/files#diff-90a04d87d7a3e373f1e17e5f80c183769c147e7034e66d1b8d6e57efe61cc150R100) and always call pytorch functions without `keepdim`. Now that both Mario and you confirmed it's indeed OK, I happily continue :-).

autogen/numpy_api_dump.py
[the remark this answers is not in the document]
ev-br: For this file specifically, I'm not sure it's worth linting: it's an autogenerated dump of the numpy namespace, made only to simplify clicking around the numpy API docs. I can take this off the repository if it gets in the way.

torch_np/_dtypes.py
```diff
@@ -72,6 +79,31 @@ def __repr__(self):
...
+}
+
+# reverse mapping
+_typecodes_from_dtype_dict = {typecode_chars_dict[key]: key
+                                for key in typecode_chars_dict}
+
+
+typecodes = {'All': 'efdFDBbhil?',
+        'AllFloat': 'efdFD',
+        'AllInteger': 'Bbhil',
```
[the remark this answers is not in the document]
ev-br: Yes, it's numpy decision. From https://numpy.org/doc/stable/reference/arrays.scalars.html (scroll down to "other types"):

> The [bool_](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.bool_) type is not a subclass of the [int_](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.int_) type (the [bool_](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.bool_) is not even a number type). This is different than Python’s default implementation of [bool](https://docs.python.org/3/library/functions.html#bool) as a sub-class of [int](https://docs.python.org/3/library/functions.html#int).

torch_np/tests/test_reductions.py
ev-br: The underlying reason is that `arg{min, max}` only accept a single axis and reject axis tuples. This is both numpy and pytorch:

```
In [176]: t
Out[176]: 
tensor([[0, 1, 2, 3, 4],
        [5, 6, 7, 8, 9]])

In [177]: t.argmin(dim=(0, 1))
TypeError                                 Traceback (most recent call last)
Input In [177], in <cell line: 1>()
----> 1 t.argmin(dim=(0, 1))

TypeError: argmin(): argument 'dim' must be int, not tuple

In [178]: t.numpy().argmin(axis=(0, 1))
TypeError                                 Traceback (most recent call last)
Input In [178], in <cell line: 1>()
----> 1 t.numpy().argmin(axis=(0, 1))

TypeError: 'tuple' object cannot be interpreted as an integer
```

Now xfailing these tests is just a technical implement: generic tests inherited from `_GenericReductionTestsMixin` test both integer axis and axis tuples. So it's either split the mixin into two (which adds complexity) or xfail in rare cases. Now that I see that it's not clear for a reader/reviewer : I'll add a comment, would that be enough? Or would `pytest.mark.skip` be clearer?

torch_np/_ndarray.py
ev-br: So you have a preference between

```
add1 = arr
if axis == ():
     arr1 = ...
```

vs

```
if axis == ():
    arr1 = ...
else:
    arr1 = arr
```
?

torch_np/_util.py
ev-br: I think I'd keep it as a utility with shapes only (thanks!) since it is used on both tensors and wrapper ndarrays. Not sure I follow the `reshape` vs `view`, could you elaborate? 
Are you saying there is an issue with `expand_dims`, https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/8/files#diff-6fff793c1370811affecb35fe3002da4f32e90d7ccf1dd2eac920c5ad86c0e02R437

torch_np/_wrapper.py
ev-br: Not sure what to note, could you elaborate? For zero-dim inputs strides seem consistent in pytorch at least (as far as strides for zero-dim arrays are actually relevant, but that's a different story)

```
In [204]: torch.argwhere(torch.empty((2,)*0, dtype=torch.int32)).stride()
Out[204]: (1, 1)

In [205]: torch.argwhere(torch.empty((2,)*0, dtype=torch.int64)).stride()
Out[205]: (1, 1)

In [206]: torch.argwhere(torch.empty((2,)*0, dtype=torch.bool)).stride()
Out[206]: (1, 1)

In [207]: _np.argwhere(_np.empty((2,)*0, dtype='int32')).strides
Out[207]: (8, 8)

In [208]: _np.argwhere(_np.empty((2,)*0, dtype='int64')).strides
Out[208]: (0, 0)

In [209]: _np.argwhere(_np.empty((2,)*0, dtype='bool')).strides
Out[209]: (8, 8)
```

ev-br:
Thank you for the review Mario! Apologies for this large PR, I hope to be able to transition to more focused PRs with time. For now e.g. the whole arange thing had to be added here since I needed something to generate arrays in tests etc.

All credit for the float16 reduction and accumulator dtype goes to numpy team, I only copy-pasted the test verbatim :-). Will need to better separate new tests from numpy clone tests, cf gh-10. 

As for generating reductions in a generic way, let's keep this in the pipeline for now. There are still several variations: ndarray methods vs free functions,  `nanmean` and relatives, also ufunc `.reduce` and `.accumulate` methods. In numpy `np.sum` delegates to `np.add.reduce`, `np.cumprod` is `np.multiply.accumulate` etc. And these are used in the wild, so maybe worth seeing if we can implement them at some point.

Before that we need to address gh-9 and gh-10 though.

autogen/numpy_api_dump.py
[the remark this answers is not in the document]
ev-br: With type hints, I'm curious on specifics of how they helped as you say, and at which stage of the lifetime of the project. I've not used them yet, and had an impression that they are more hassle than worth.

torch_np/_ndarray.py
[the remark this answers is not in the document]
ev-br: All credit goes to the numpy team, I've only copy-pasted the test.

autogen/numpy_api_dump.py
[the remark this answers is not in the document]
ezyang: Black formatting is pretty cheap and better to do earlier in the lifetime of the project. Typing I am a bit more ambivalent about; it works well if you are willing to put in the effort and have complete control over API design, but a lot of Pythonic API style isn't easily typeable and you have to be willing to not do it, or put up with suppressions

torch_np/_ndarray.py
lezcano: My point is, why not doing this using the `arr` name itself. Then simply do `arr = arr.reshape(newshape)`

autogen/numpy_api_dump.py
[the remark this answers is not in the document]
lezcano: About how type hints helped: it was helpful to express the semantics of the functions that we had. It made the code much easier to read and reason about. It also exposed bugs early. I mean, like any other type system really.

It's also true that it takes a bit of effort to maintain, as Ed mentioned. 

In any case, I'm +1 on black and +0 on the type system.

torch_np/tests/test_reductions.py
```diff
@@ -0,0 +1,617 @@
...
+            for axis in [None]:
+                tgt = mat.sum(axis=axis)
+                res = np.mean(mat, axis=axis) * mat.size
+                assert_allclose(res, tgt)
+
+    @pytest.mark.xfail(reason="XXX: mean(..., dtype=..) smarter dtype selection")
+    def test_mean_float16(self):
+        # This fail if the sum inside mean is done in float16 instead
+        # of float32.
+        assert np.mean(np.ones(100000, dtype='float16')) == 1
```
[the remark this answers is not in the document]
lezcano: See also the responses there about float16 and CPU support on PyTorch.

torch_np/tests/test_reductions.py
lezcano: I like `xfail`s for bugs and `skip`s for things that we know that fail and we accept that. That way, if at some point we fix the given bug CI will remind us to remove those xfails. What threw me off was the word `axis` in the xfail comment, as in PyTorch we call it `dim` or `dims`.

About this particular test, I see that both implementations error out on tuples, so I still don't understand what's the difference between the two, sorry :/

torch_np/_wrapper.py
lezcano: For larger inputs, the strides returned by this implementation will be transposed, while I'd expect those from NumPy to be contiguous.

autogen/numpy_api_dump.py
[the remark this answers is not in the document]
rgommers: +1 for adding `black` too, at some point before merging code where it needs to be maintained long term.

> it works well if you are willing to put in the effort and have complete control over API design, but a lot of Pythonic API style isn't easily typeable and you have to be willing to not do it, or put up with suppressions

Indeed. When starting from scratch, type annotations are great - exactly because they help you avoid problematic constructs (the kinds that aren't easy to statically type, and those NumPy has a ton of). For this project we have to adhere to the NumPy API, and we can't really avoid array-like inputs, dtype keywords that take multiple kinds of types, polymorphic functions like `unique` and `cross`, etc. 

So here, I'd be neutral at best on type annotations. At least as far as exposing them goes - if they're internal only and some shortcuts can be taken when the above-mentioned constructs are encountered, then they may still be helpful.

torch_np/tests/test_reductions.py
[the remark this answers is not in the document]
ev-br: While I don't have much to say on pytorch policies,  I'd also expect this to be classified as a bug not an enhancement. 

Meanwhile, here the fix is simple
https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/8/commits/01762f6b6e4f9f7a41aa19ec5a6c2e81f8b5fa4a
I'm going to postpone generalizing this from `mean` to other functions until there's testing.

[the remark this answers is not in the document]
lezcano: yup, but I don't think we should work around PyTorch bugs. It's better to xfail them and report them. When someone fixes them in master, we can remove the relevant xfails.

torch_np/tests/test_reductions.py
ev-br: OK so the xfails are actually not needed: the tests were in fact xpassing. These tests check that basically,
- `reduction_func(a, axis=(1, 1))` errors out on a repeated axis
- `reduction_func(a, axis=())` behaves as expected.

Now both behaviors are handled before handling the heavy lifting to pytorch, so both behaviors are consistent with numpy now (in a couple of minutes when I push).

torch_np/_ndarray.py
ev-br: OK, done.

torch_np/_util.py
ev-br: ISTM numpy.reshape semantics is also to return a view if possible, a copy otherwise:

> This will be a new view object if possible; otherwise, it will be a copy. Note there is no guarantee of the memory layout (C- or Fortran- contiguous) of the returned array. 

https://numpy.org/doc/stable/reference/generated/numpy.reshape.html

I would be *very* surprised if numpy is ever raising on a reshape as long as the number of array elements is the same.

Or are you saying that the last line of

```
def expand_dims(a, axis):
    a = asarray(a)
    shape = _util.expand_shape(a.shape, axis)
    return a.reshape(shape)
```

should read `return a.view(shape)` instead or `return a.reshape(shape)`?

torch_np/_wrapper.py
ev-br: OK,  with `argwhere` just delegating to `pytorch.argwhere`, it all seems consistent (modulo byte vs element strides):

```
In [133]: t
Out[133]: 
tensor([[0., 1., 2., 3., 4.],
        [5., 6., 7., 8., 9.]])

In [134]: w.asarray(t)
Out[134]: 
array_w([[0., 1., 2., 3., 4.],
        [5., 6., 7., 8., 9.]])

In [135]: w.argwhere(w.asarray(t))
Out[135]: 
array_w([[0, 1],
        [0, 2],
        [0, 3],
        [0, 4],
        [1, 0],
        [1, 1],
        [1, 2],
        [1, 3],
        [1, 4]])

In [136]: w.argwhere(w.asarray(t)).strides
Out[136]: (1, 9)

In [137]: np.argwhere(t.numpy()).strides
Out[137]: (8, 72)
```

torch_np/tests/test_reductions.py
lezcano: how come you removed the whole test, rather than just hte xfails?

torch_np/_util.py
lezcano: Yes, basically to signal that it will never copy. We wouldn't want this function to copy, as we're just adding a few dimensions with legnth 1.

torch_np/tests/test_reductions.py
ev-br: Nope, the test is there through the inheritance:

```
class _GenericReductionsTestMixin:        # class name does not start with Test, pytest skips it
    def test_axis_bad_tuple(self):
        # Basic check of functionality
        m = np.array([[0, 1, 7, 0, 0], [3, 0, 0, 2, 19]])
        with assert_raises(ValueError):
             self.func(m, axis=(1, 1))                   # note self.func

class TestAnyGeneric(_GenericReductionsTestMixin):      # class name starts with Test
    def setup_method(self):
        self.func = np.any
```

I see that the whole xfail story was confusing, but it is gone now anyway :-).

torch_np/_util.py
ev-br: OK, done: https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/8/commits/2da3136b2835b9d6a6016b30dfb154c016fe0776

In numpy-speak `ndarray.view` is IIUC is only about dtypes not shapes, which is why it took to this long to get to what you were saying. Anyway, as long as the main consumer here is pytorch itself., let's make it easier for pytorch :-).

ev-br:
OK, I think all review comments are addressed, modulo (i)a more generic implementation of all reductions --- which I believe is best postponed for a while, and (ii) using black --- which also needs to be done in one go on main.
Any further comments, or does anybody want more time to review this? 
(Once merged, this is all visible in gh-2)

autogen/numpy_api_dump.py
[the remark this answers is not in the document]
ev-br: One maintenancy consideration of blackening (is that a word?) is gh-10. tests vendored from numpy should likely by excluded to minimize the diff size going forward.

torch_np/_util.py
lezcano: Oh, I see now. Sorry, my NumPy is not as good as it could be, and this starts to feel like talking about the differences between C and C++...

lezcano:
I'll give it another pass tomorrow, but at first sight it looks good!

torch_np/_ndarray.py
```diff
@@ -120,8 +176,27 @@ def squeeze(self, axis=None):
...
+        return tensor
+
+    @axis_out_keepdims_wrapper
+    def argmin(self, axis=None, out=None, *, keepdims=NoValue):
+        axis = _helpers.allow_only_single_axis(axis)
+
+        if axis is None:
+            tensor = torch.argmin(self._tensor)
+        else:
+            tensor = torch.argmin(self._tensor, axis)
```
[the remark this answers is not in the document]
lezcano: Both argmin and argmax

torch_np/_wrapper.py
```diff
@@ -179,14 +177,28 @@ def logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None, axis=0):
     return asarray(torch.logspace(start, stop, num, base=base, dtype=dtype))
 
 
-def arange(start, stop=None, step=1, dtype=None, *, like=None):
+def arange(start=None, stop=None, step=1, dtype=None, *, like=None):
```
lezcano, line 180: Why not setting `start=0`

lezcano (approved):
LGTM. I left a number of minor comments, but these shouldn't be blocking.

torch_np/_ndarray.py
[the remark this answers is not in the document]
ev-br: Thanks! It is not indeed, changed. Speaking of minor inconsistensies, it is needed for squeeze:

```
In [262]: t.squeeze(axis=None)
Input In [262], in <cell line: 1>()
----> 1 t.squeeze(axis=None)

RuntimeError: Please look up dimensions by name, got: name = None.
```

torch_np/_ndarray.py
```diff
@@ -95,19 +132,38 @@ def __float__(self):
...
+        return asarray(self._tensor.__sub__(other_tensor))
 
     def __mul__(self, other):
-        return self._tensor.__mul__(other)
+        other_tensor = asarray(other).get()
+        return asarray(self._tensor.__mul__(other_tensor))
+
+    def __rmul__(self, other):
+        other_tensor = asarray(other).get()
+        return asarray(self._tensor.__mul__(other_tensor))
```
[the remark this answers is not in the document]
ev-br: Thanks, addressed. These are untested and unfinished yet.

torch_np/_wrapper.py
ev-br: I did not manage to make all numpy tests pass otherwise. No doubt is can be fixed, but I'd rather move it to "not a first prio" bucket if you don't mind.

torch_np/_ndarray.py
```diff
@@ -3,11 +3,39 @@
...
 newaxis = None
 
+
+def axis_out_keepdims_wrapper(func):
+    """`func` accepts an array-like as a 1st arg, returns a tensor.
+
+    This decorator implements the generic handling of axis, out and keepdims
+    arguments for reduction functions.
+    """
+    # XXX: move this out of _ndarray.py (circular imports)
```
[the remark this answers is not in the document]
ev-br: At some point, certainly. For now what needs what is still in flux, so I'd rather flag it for some point in the future and move on. For the moment, this decorator needs `asarray`, which knows about `ndarray` internals, and the decorator itself is used for `ndarray` members :-).

ev-br:
Thanks Mario for the review. 
Since there are no known blockers and the CI shows the same expected 10 failures from bad tests, let's get this in and iterate.

Issue #9, opened 2023-01-02
Dtypes


ev-br:
The current dtype stubs are too simplistic. Let's see if we can make them work without reproducing the whole story about dtypes [1] vs array scalars [2].  
The minimum set of requirements is
- `np.float32(42)` should work. In numpy it creates a scalar, can we get around creating a single-element zero-dimensional array (backed by a zero-dimensional tensor).
- (a part of) dtype class hierarchy is needed. This includes common patterns like `if np.issubdtype(x.dtype, np.complexfloating)`. Sadly, there are also `isinstance` checks in the numpy test suite:

```
(numpy-dev) br@gonzales:~/repos/numpy$ grep -rn np.inexact .
./doc/source/release/1.19.0-notes.rst:259:or ``np.inexact`` will now give a deprecation warning when converted
./doc/neps/nep-0040-legacy-datatype-impl.rst:254:types such as ``np.inexact`` (see figure below).
./doc/neps/nep-0040-legacy-datatype-impl.rst:256:``issubclass(a.dtype.type, np.inexact)``.
./doc/neps/nep-0042-new-dtypes.rst:330:   ``isinstance(np.dtype("float64"), np.inexact)``.
./numpy/random/bit_generator.pyx:139:    elif isinstance(x, (float, np.inexact)):
./numpy/linalg/tests/test_linalg.py:1229:        if issubclass(x.dtype.type, np.inexact):
./numpy/lib/function_base.py:1208:    elif np.issubdtype(otype, np.inexact):
./numpy/lib/function_base.py:3876:    if np.issubdtype(a.dtype, np.inexact):
./numpy/lib/function_base.py:3908:    if np.issubdtype(a.dtype, np.inexact) and sz > 0:
./numpy/lib/function_base.py:4681:    if np.issubdtype(arr.dtype, np.inexact):
./numpy/lib/function_base.py:4733:        if np.issubdtype(arr.dtype, np.inexact):
./numpy/lib/function_base.py:4753:        if np.issubdtype(arr.dtype, np.inexact):
./numpy/lib/nanfunctions.py:101:    elif issubclass(a.dtype.type, np.inexact):
./numpy/lib/nanfunctions.py:1039:    if dtype is not None and not issubclass(dtype.type, np.inexact):
./numpy/lib/nanfunctions.py:1041:    if out is not None and not issubclass(out.dtype.type, np.inexact):
[...]

Issue #5, opened 2022-12-16
Tracker: semantic differences between numpy and pytorch


ev-br:
This is to track situations where eponymous functions in numpy and torch have different semantics, and to record decisions about these. Typically, the difference is views vs copies.

- flip: `np.flip` creates a view, `torch.flip` creates a copy.

ev-br:
Here's one possibly ugly corner with scalars vs zero-dim arrays. I'm trying to have a system where `tnp.float32(2)` creates a zero-dim array which wraps `torch.Tensor(2, dtype=torch.float32)`. A possible issue is what `tnp.int32(2) * [1, 2, 3]` do: 

```
In [353]: np.int32(2) * [1, 2, 3]          # scalar decays to a python int
Out[353]: [1, 2, 3, 1, 2, 3]

In [357]: np.asarray(2) * [1, 2, 3]     # zero-dim array is an array-like
Out[357]: array([2, 4, 6])

```

rgommers:
Hmm, maybe it's necessary to not support using as Python scalars (e.g. by using an ndarray subclass or a special flag to indicate that the 0-D array was actually a scalar). Or maybe it's too much of a corner case. Probably for now just document the problem, add an `xfail` test, and move on?

ev-br:
Let's see if we can pull off the following: scalars are 0d arrays and do not decay to python scalars unless explicitly requested. In the example above this means the Out[357] behavior unless there is an explicit conversion to the python scalar, as in int(np.int32(2)) * [1, 2, 3]. We might even follow pytorch in allowing int(torch.Tensor([[[1]]])) which numpy raises on.

ev-br:
As long as we do not special-case scalars, let's relax the NumPy casting rule that scalars do not upcast arrays. So, the casting rule is then everything upcasts everything else, and there is no value-based casting.

ev-br:
Am splitting this tracker into two: this one is for numpy/pytorch differences, and gh-73 attempt to track the differences between the original numpy and our wrapper.

PR #12, opened 2023-01-03
dtypes and arrays scalars

ev-br:
Address gh-9: implement the scalar type and array dtype hierarchy.

Implementation-wise it does a fairly weird thing of overriding `__new__` so that e.g.
`np.float32(1)` constructs a zero-dim tensor, *not*  a `np.float32` instance. 

See also https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/5#issuecomment-1371860957 and below for a discussion about array scalars.

closes gh-9 when done.


lezcano:
nit. Why don't you submit this as a merge into `reductions` to get a cleaner diff in the whole PR?

ev-br:
Note that the initial review is at gh-13.

ev-br:
OK, this works up to `test_scalarmath`. 
Fixing that one involves implementing a decent subset of ufuncs, so I'm going to take it to a separate branch.

.gitignore
```diff
@@ -3,5 +3,6 @@ autogen/__pycache__
 torch_np/__pycache__/*
 torch_np/tests/__pycache__/*
 torch_np/tests/numpy_tests/core/__pycache__/*
+torch_np/testing/__pycache__/*
```
lezcano, line 6:
```suggestion
**/__pycache__/
```
Let's just ignore them all.

lezcano: Even more, a better syntax may be
```python
# Byte-compiled / optimized / DLL files
__pycache__/
*.py[cod]
```
as per the [oficial gitignore](https://github.com/github/gitignore/blob/main/Python.gitignore).

ev-br:
Ok, thanks for the reviews and discussions, let's merge this and iterate.

PR #13, opened 2023-01-03
dtypes and array scalars

ev-br:
identical to gh-12, only sent to the _reductions_ branch, as per https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/12#issuecomment-1369750928


asmeurer:
I guess there's a lot of other methods on numpy dtypes that people might care about (although none of them are guaranteed by the array API FWIW). There's also the extensive `__eq__` that is defined on numpy dtypes (see also https://github.com/data-apis/array-api/issues/582). NumPy also has the distinction between dtype classes and instances (e.g., `np.float64` vs. `np.dtype('float64')`, which I personally find annoying and I don't know if it needs to exist here.

torch_np/_dtypes.py
```diff
@@ -183,6 +181,23 @@ def is_integer(dtyp):
...
+        return False
+
+
+def issubdtype(arg1, arg2):
+    # cf https://github.com/numpy/numpy/blob/v1.24.0/numpy/core/numerictypes.py#L356-L420
+    if not issubclass_(arg1, _scalar_types.generic):
+        arg1 = dtype(arg1).type
+    if not issubclass_(arg2, _scalar_types.generic):
+        arg2 = dtype(arg2).type
+    return issubclass(arg1, arg2)
```
rgommers, line 198: This seems to not yet be complete I think? At least from a quick test:
```python
In [25]: tnp.issubdtype(tnp.float32, tnp.inexact)
Out[25]: False

In [26]: np.issubdtype(np.float32, np.inexact)
Out[26]: True
```

torch_np/_dtypes.py
```diff
@@ -183,6 +181,23 @@ def is_integer(dtyp):
     return dtyp.typecode in typecodes['AllInteger']
 
 
+
+def issubclass_(arg, klass):
```
rgommers, line 185: This goes on my list of functions to deprecate and remove in NumPy 2.0 ...

rgommers: This one is easy enough so may as well leave it here now. I'd say let's not bother with `issubsctype`, `obj2sctype`, etc., those definitely all need to go from NumPy.

rgommers (commented):
This looks like a good approach overall. I think it will work well enough. It would be useful to add the main properties that are supported in a docstring (probably `generic`, because that's where the functionality is).

> There's also the extensive `__eq__` that is defined on numpy dtypes (see also https://github.com/data-apis/array-api/issues/582). 

`__eq__` would indeed be good to support as much as possible. 

> NumPy also has the distinction between dtype classes and instances (e.g., `np.float64` vs. `np.dtype('float64')`, which I personally find annoying and I don't know if it needs to exist here. 

Agreed it's annoying, but it's probably easier to keep that here. The correspondence in this PR isn't perfect (see example below), but close enough.

```python
>>> type(tnp.dtype('float64'))
<class 'torch_np._dtypes.dtype'>
>>> type(np.dtype('float64'))
<class 'numpy.dtype[float64]'>
```

PR #16, opened 2023-01-06
port scalarmath tests

ev-br:
Make minimal modifications to dtype/scalar type machinery to get scalar math tests run. Lots of xfails/test skips for now, but I think this is more or less it for dtypes. This PR is against the gh-12 branch, these two PRs are a single "stack", as it were.

Next up is ufuncs, esp those which are also ndarray methods.

### DIFF
--- a/torch_np/_ndarray.py
+++ b/torch_np/_ndarray.py
@@ -81,6 +81,18 @@ class ndarray:
     def T(self):
         return self.transpose()
 
+    @property
+    def real(self):
+        return asarray(self._tensor.real)
+
+    @property
+    def imag(self):
+        try:
+            return asarray(self._tensor.imag)
+        except RuntimeError:
+            zeros = torch.zeros_like(self._tensor)
+            return ndarray._from_tensor_and_base(zeros, None)
+
     # ctors
     def astype(self, dtype):
         newt = ndarray()
@@ -102,6 +114,13 @@ class ndarray:
 
     ### comparisons ###
     def __eq__(self, other):
+        try:
+            t_other = asarray(other).get
+        except RuntimeError:
+            # Failed to convert other to array: definitely not equal.
+            # TODO: generalize, delegate to ufuncs
+            falsy = torch.full(self.shape, fill_value=False, dtype=bool)
+            return asarray(falsy)
         return asarray(self._tensor == asarray(other).get())
 
     def __neq__(self, other):
@@ -118,7 +137,6 @@ class ndarray:
 
     def __le__(self, other):
         return asarray(self._tensor <= asarray(other).get())
-
 
     def __bool__(self):
         try:
@@ -167,7 +185,10 @@ class ndarray:
 
     def __sub__(self, other):
         other_tensor = asarray(other).get()
-        return asarray(self._tensor.__sub__(other_tensor))
+        try:
+            return asarray(self._tensor.__sub__(other_tensor))
+        except RuntimeError as e:
+            raise TypeError(e.args)
 
     def __mul__(self, other):
         other_tensor = asarray(other).get()
@@ -177,9 +198,29 @@ class ndarray:
         other_tensor = asarray(other).get()
         return asarray(self._tensor.__rmul__(other_tensor))
 
+    def __floordiv__(self, other):
+        other_tensor = asarray(other).get()
+        return asarray(self._tensor.__floordiv__(other_tensor))
+
+    def __ifloordiv__(self, other):
+        other_tensor = asarray(other).get()
+        return asarray(self._tensor.__ifloordiv__(other_tensor))
+
     def __truediv__(self, other):
         other_tensor = asarray(other).get()
         return asarray(self._tensor.__truediv__(other_tensor))
+
+    def __itruediv__(self, other):
+        other_tensor = asarray(other).get()
+        return asarray(self._tensor.__itruediv__(other_tensor))
+
+    def __mod__(self, other):
+        other_tensor = asarray(other).get()
+        return asarray(self._tensor.__mod__(other_tensor))
+
+    def __imod__(self, other):
+        other_tensor = asarray(other).get()
+        return asarray(self._tensor.__imod__(other_tensor))
 
     def __or__(self, other):
         other_tensor = asarray(other).get()
@@ -189,9 +230,21 @@ class ndarray:
         other_tensor = asarray(other).get()
         return asarray(self._tensor.__ior__(other_tensor))
 
-
     def __invert__(self):
         return asarray(self._tensor.__invert__())
+
+    def __abs__(self):
+        return asarray(self._tensor.__abs__())
+
+    def __neg__(self):
+        try:
+            return asarray(self._tensor.__neg__())
+        except RuntimeError as e:
+            raise TypeError(e.args)
+
+    def __pow__(self, exponent):
+        exponent_tensor = asarray(exponent).get()
+        return asarray(self._tensor.__pow__(exponent_tensor))
 
     ### methods to match namespace functions
 
--- a/torch_np/_ndarray.py
+++ b/torch_np/_ndarray.py
@@ -159,6 +159,9 @@ class ndarray:
     def __float__(self):
         return float(self._tensor)
 
+    def __int__(self):
+        return int(self._tensor)
+
     # XXX : are single-element ndarrays scalars?
     def is_integer(self):
         if self.shape == ():
@@ -354,7 +357,7 @@ class ndarray:
 
         if dtype is None:
             dtype = self.dtype
-        if not _dtypes.is_floating(dtype):
+        if _dtypes.is_integer(dtype):
             dtype = _dtypes.default_float_type()
         torch_dtype = _dtypes.torch_dtype_from(dtype)
 
@@ -374,7 +377,7 @@ class ndarray:
 
         if dtype is None:
             dtype = self.dtype
-        if not _dtypes.is_floating(dtype):
+        if _dtypes.is_integer(dtype):
             dtype = _dtypes.default_float_type()
         torch_dtype = _dtypes.torch_dtype_from(dtype)
 
@@ -396,36 +399,8 @@ class ndarray:
         return self._tensor.__setitem__(index, value)
 
 
-def asarray(a, dtype=None, order=None, *, like=None):
-    _util.subok_not_ok(like)
-    if order is not None:
-        raise NotImplementedError
-
-    if isinstance(a, ndarray):
-        if dtype is not None and dtype != a.dtype:
-            a = a.astype(dtype)
-        return a
-
-    if isinstance(a, (list, tuple)):
-        # handle lists of ndarrays, [1, [2, 3], ndarray(4)] etc
-        a1 = []
-        for elem in a:
-            if isinstance(elem, ndarray):
-                a1.append(elem.get().tolist())
-            else:
-                a1.append(elem)
-    else:
-        a1 = a
-
-    torch_dtype = _dtypes.torch_dtype_from(dtype)
-
-    # This and array(...) are the only places which talk to ndarray directly.
-    # The rest goes through asarray (preferred) or array.
-    out = ndarray()
-    tt = torch.as_tensor(a1, dtype=torch_dtype)
-    out._tensor = tt
-    return out
-
+# This is the ideally the only place which talks to ndarray directly.
+# The rest goes through asarray (preferred) or array.
 
 def array(object, dtype=None, *, copy=True, order='K', subok=False, ndmin=0,
           like=None):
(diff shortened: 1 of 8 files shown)


torch_np/_ndarray.py
```diff
@@ -81,6 +81,18 @@ def base(self):
...
+    def real(self):
+        return asarray(self._tensor.real)
+
+    @property
+    def imag(self):
+        try:
+            return asarray(self._tensor.imag)
+        except RuntimeError:
+            zeros = torch.zeros_like(self._tensor)
+            return ndarray._from_tensor_and_base(zeros, None)
```
lezcano, line 94: Why not simply `return asarray(zeros)`?

torch_np/_ndarray.py
```diff
@@ -102,6 +114,13 @@ def __str__(self):
 
     ### comparisons ###
     def __eq__(self, other):
+        try:
+            t_other = asarray(other).get
+        except RuntimeError:
+            # Failed to convert other to array: definitely not equal.
+            # TODO: generalize, delegate to ufuncs
+            falsy = torch.full(self.shape, fill_value=False, dtype=bool)
+            return asarray(falsy)
```
lezcano, line 123: I don't think this is correct
```python
>>> np.ndarray([5]) == [3, 4, 5]
<stdin>:1: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
False
```

torch_np/_ndarray.py
```diff
@@ -167,7 +188,10 @@ def __iadd__(self, other):
 
     def __sub__(self, other):
         other_tensor = asarray(other).get()
-        return asarray(self._tensor.__sub__(other_tensor))
+        try:
+            return asarray(self._tensor.__sub__(other_tensor))
+        except RuntimeError as e:
+            raise TypeError(e.args)
```
lezcano, line 194: Why here and not everywhere else?

torch_np/_ndarray.py
```diff
@@ -189,10 +233,22 @@ def __ior__(self, other):
...
         return asarray(self._tensor.__invert__())
 
+    def __abs__(self):
+        return asarray(self._tensor.__abs__())
+
+    def __neg__(self):
+        try:
+            return asarray(self._tensor.__neg__())
+        except RuntimeError as e:
+            raise TypeError(e.args)
```
lezcano, line 246: Same as above...

torch_np/_ndarray.py
```diff
@@ -343,67 +399,80 @@ def __setitem__(self, index, value):
         return self._tensor.__setitem__(index, value)
 
 
-def asarray(a, dtype=None, order=None, *, like=None):
-    _util.subok_not_ok(like)
-    if order is not None:
+# This is the ideally the only place which talks to ndarray directly.
+# The rest goes through asarray (preferred) or array.
+
+def array(object, dtype=None, *, copy=True, order='K', subok=False, ndmin=0,
```
lezcano, line 405: nit. `object` is a builtin class in Python, perhaps use a different name?

torch_np/_ndarray.py
```diff
@@ -343,67 +399,80 @@ def __setitem__(self, index, value):
...
+          like=None):
+    _util.subok_not_ok(like, subok)
+    if order != 'K':
```
lezcano, line 408: refactor into a `_util` function?

torch_np/_ndarray.py
```diff
@@ -343,67 +399,80 @@ def __setitem__(self, index, value):
...
-    tt = torch.as_tensor(a1, dtype=torch_dtype)
-    out._tensor = tt
-    return out
+        # int(torch.as_tensor(1e12)) - 1e12 equals -4096 (try it!)
 
+        # Therefore, we treat `tensor.dtype` as a hint, and convert the
+        # original object *again*, this time with an explicit dtype.
+        dtyp = _dtypes.dtype_from_torch(tensor.dtype)
+        default = _dtypes.get_default_dtype_for(dtyp)
+        torch_dtype = _dtypes.torch_dtype_from(default)
```
lezcano, line 448: we may want to overload `get_default_dtype_for` to also accept PyTorch dtypes to simplify this code.

lezcano: Also, I get the feeling that if PyTorch defaults are the same as those in NumPy (via `set_default...`) simply doing `as_tensor` should do the trick. Is this possible?
Otherwise, as you mention above, we should make sure we just call `as_tensor`  from here, as otherwise we'll inadvertently get incorrect results.

torch_np/_ndarray.py
```diff
@@ -343,67 +399,80 @@ def __setitem__(self, index, value):
...
-        if dtype != object.dtype:
-            torch_dtype = _dtypes.torch_dtype_from(dtype)
-            result = result.to(torch_dtype)
-    else:
+    # type cast if requested
+    if dtype is not None:
         torch_dtype = _dtypes.torch_dtype_from(dtype)
-        result = torch.as_tensor(object, dtype=torch_dtype)
+        tensor = tensor.to(torch_dtype)
+        base = None
```
lezcano, line 456: Note. This is alright because casting functions will be fused with creation functions, but in PyTorch eager would be better to simply use this dtype in the `as_tensor`s calls above.

torch_np/tests/numpy_tests/core/test_scalarmath.py
```diff
@@ -302,7 +306,7 @@ def test_float_modulus_exact(self):
     def test_float_modulus_roundoff(self):
         # gh-6127
         dt = np.typecodes['Float']
-        for op in [floordiv_and_mod, divmod]:
+        for op in [floordiv_and_mod]:  # TODO divmod is not implemented
```
lezcano, line 309: assert it below so that it starts failing when we implement it or simply xfail it as the previous test.

torch_np/tests/numpy_tests/core/test_scalarmath.py
```diff
@@ -352,18 +357,10 @@ def test_float_modulus_corner_cases(self):
                     div, mod = op(fone, fzer)
                     assert_(np.isinf(div)) and assert_(np.isnan(mod))
 
-    def test_inplace_floordiv_handling(self):
```
lezcano, line 355: why removed?

torch_np/tests/numpy_tests/core/test_scalarmath.py
```diff
@@ -352,18 +357,10 @@ def test_float_modulus_corner_cases(self):
...
-        a = np.array([1, 2], np.int64)
-        b = np.array([1, 2], np.uint64)
-        with pytest.raises(TypeError,
-                match=r"Cannot cast ufunc 'floor_divide' output from"):
-            a //= b
-
 
 class TestComplexDivision:
+
+    @pytest.mark.xfail(reason='With pytorch, 1/(0+0j) is nan + nan*j, not inf + nan*j')
```
lezcano, line 363: probably simply skip. These are the sort of incongruences we should be alright with having.

torch_np/tests/numpy_tests/core/test_scalarmath.py
```diff
@@ -437,22 +434,24 @@ def test_branches(self):
 
 class TestConversion:
     def test_int_from_long(self):
+        # NB: this test assumes that the default fp type is float64
```
lezcano, line 437: which we have now! nice.

torch_np/tests/numpy_tests/core/test_scalarmath.py
```diff
@@ -437,22 +434,24 @@ def test_branches(self):
...
         l = [1e6, 1e12, 1e18, -1e6, -1e12, -1e18]
         li = [10**6, 10**12, 10**18, -10**6, -10**12, -10**18]
         for T in [None, np.float64, np.int64]:
             a = np.array(l, dtype=T)
             assert_equal([int(_m) for _m in a], li)
 
-        a = np.array(l[:3], dtype=np.uint64)
-        assert_equal([int(_m) for _m in a], li[:3])
```
lezcano, line 447: why are these deleted?

torch_np/tests/numpy_tests/core/test_scalarmath.py
```diff
@@ -483,13 +482,13 @@ def test_numpy_scalar_relational_operators(self):
...
         #Unsigned integers
-        for dt1 in 'BHILQP':
+        for dt1 in 'B':
             assert_(-1 < np.array(1, dtype=dt1)[()], "type %s failed" % (dt1,))
             assert_(not -1 > np.array(1, dtype=dt1)[()], "type %s failed" % (dt1,))
             assert_(-1 != np.array(1, dtype=dt1)[()], "type %s failed" % (dt1,))
 
             #unsigned vs signed
-            for dt2 in 'bhilqp':
+            for dt2 in 'bhil':
```
lezcano, line 491: nb. All these tiny changes of things we don't support would be better factored out into their own PR to make reviews easier.

torch_np/tests/numpy_tests/core/test_scalarmath.py
```diff
@@ -703,6 +680,8 @@ def test_builtin_abs(self, dtype):
     def test_numpy_abs(self, dtype):
         self._test_abs_func(np.abs, dtype)
 
+
+@pytest.mark.skip(reason='TODO: implement bit shifts')
```
lezcano, line 684: xfail?

lezcano (commented):
A general question. Can you remind me why we need to explicitly have a `base` member within `_ndarray`? What use cases does it cover?

ev-br:
> A general question. Can you remind me why we need to explicitly have a base member within _ndarray? What use cases does it cover?

In numpy it's, AFAIU, a python side of the pointer chain for ownership of the array data:

```
In [1]: import numpy as np

In [2]: a = np.arange(4)

In [3]: a[::2].base is a
Out[3]: True

In [4]: a[::2][1] = 8

In [5]: a
Out[5]: array([0, 1, 8, 3])
```

So if `arr` is a view of some other array, `a.base` points to it. (Here `view` is numpy view, not pytorch'es).
Since it's public in numpy, why would we not keep it as well.

torch_np/_ndarray.py
ev-br: This circles back to an earlier discussion of who's the main intended consumer. 

E.g., emulation of `out=` arguments is certainly suboptimal for pytorch eager, as pytorch functions never receive the `out` tensors. It does make life easier for the wrapper though. 

So what do we aim to, and what is the balance between efficiency and simplicity we're aiming for.

torch_np/_ndarray.py
ev-br: Let's maybe not rely on this, not just yet at least. When we have a more complete coverage, let's experiment with how much we can peel out.

torch_np/_ndarray.py
ev-br: Definitely. In a separate PR a bit later down the line, I'd say. Across the whole codebase

torch_np/_ndarray.py
ev-br: Yeah. So far I matched numpy names one to one, but this one is indeed unfortunate.

torch_np/_ndarray.py
ev-br: This is gone in the next PRs in this stack.

torch_np/_ndarray.py
ev-br: At some point we'll need to rationalize these two forms, agree.
Basically, `asarray` is anything array-like in, array out; here we explicitly construct the tensor, so my fingers naturally typed this line. The line above, with `asarray(self._tensor.imag)`, should be changed to follow line 94.

torch_np/_ndarray.py
ev-br: Hm. I thought that using `np.ndarray` was merely not recommended as an array constructor, but did not realize that it does behave *this* differently from a regular `np.array`:

```
In [12]: np.ndarray([6]) == [1, 2, 3]
<ipython-input-12-f4f66d3e4ff6>:1: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
  np.ndarray([6]) == [1, 2, 3]
Out[12]: False

In [13]: np.array([6]) == [1, 2, 3]                  # np.array, not *nd*array
Out[13]: array([False, False, False])

In [14]: np.array([6]) == 'oops'
<ipython-input-14-0ccc45613206>:1: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  np.array([6]) == 'oops'
Out[14]: False
```

I think this because the first argument of `np.ndarray` is `shape`: https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html

```
In [19]: np.ndarray([5])
Out[19]: 
array([4.65458471e-310, 0.00000000e+000, 4.94065646e-324,             nan,
       6.92900938e-310])
```

So your example is essentially `np.empty(5) == [1, 2, 3]` and that rightly fails, while `np.ndarray([5]) == [1, 2, 3, 4, 5]` gives an array of five `False`s.

ev-br: That said, this specific try-catch intends to catch a more common issue: consider `np.array([3]) == "oops"`. Since we don't have object arrays, `asarray(other)` fails and we need to explicitly emulate what NumPy's FutureWarning promises. Since `other` is not an array-like, out best bet is to treat it as a scalar, which gives a False on a comparison to each element of `self`.

torch_np/tests/numpy_tests/core/test_scalarmath.py
ev-br: This test is AFAU specific to combining signed and unsigned 64-bit integers

```
        # this only applies to in-place floordiv //=, because the output type
        # promotes to float which does not fit
        a = np.array([1, 2], np.int64)
        b = np.array([1, 2], np.uint64)
        with pytest.raises(TypeError,
                match=r"Cannot cast ufunc 'floor_divide' output from"):
            a //= b
```

Chainging the dtype of b does not cause any trouble:

```
In [13]: a = np.array([1, 2], np.int64) ; a //= b.astype('uint8'); a
Out[13]: array([1, 1])

In [14]: a = np.array([1, 2], np.int64) ; a //= b.astype('int32'); a
Out[14]: array([1, 1])

In [15]: a = np.array([1, 2], np.int64) ; a //= b.astype('int64'); a
Out[15]: array([1, 1])

```

torch_np/tests/numpy_tests/core/test_scalarmath.py
ev-br: It replicates the test above with `uint64`.

ev-br:
Review comments addressed in https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/16/commits/2b95ac252dd48cb915d18fd7fd91a75e33318aa5

torch_np/_ndarray.py
lezcano: I mean, `asarray(Tensor)` has the same semantics as `_from_tensor_and_base(Tensor, None)`, so we can decide to always prefer the first one over the latter one for conciseness and consistency.

torch_np/_ndarray.py
lezcano: About my example, yep, I know it creates a tensor of 5 elements. My point is that the current code never returns `False`, while NumPy may return a straight `False` from `__eq__`. Another example is the `"oops"` you've written there, which also returns `False`, not a tensor full of `False`s.

torch_np/_ndarray.py
lezcano: It's probably fine as it is. Let's optimise for a clean codebase for now.

torch_np/_ndarray.py
ev-br: Not always

```
In [2]: a = np.zeros(3)

In [3]: np.asarray(a) is a
Out[3]: True
```

torch_np/tests/numpy_tests/core/test_scalarmath.py
lezcano: Right. As discussed, moving forward, it'd be good to put together all these test removals into one PR that can be merged with little review.

lezcano:
About https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/16#issuecomment-1379117385, sure, but this tracking also happens in `torch.tensor()._base`, so why not just return that one?

torch_np/_ndarray.py
lezcano: note that I suggested doing so when using `torch.Tensor`s.

torch_np/_ndarray.py
ev-br: Yes, the current version here follows the "but in the future will perform elementwise comparison" from the deprecation message. With the "NumPy 2.0" discussion in the numpy ML, it looks like the timescale for the future is about a year.

lezcano: Ah, right, fair.

lezcano (approved):
We should discuss `asarray` vs `_from_tensor_and_base` (probably remove the latter one) and potentially removing the `base` attribute from the wrapper. Now, we don't need to block this PR on that.

ev-br:
Thanks for the review Mario!
Merged in https://github.com/Quansight-Labs/numpy_pytorch_interop/commit/f63c6bbf7305d7093e4a008328042e61e47c8cb9

PR #17, opened 2023-01-08
ndarray dunders  / binary ufuncs

ev-br:
This is a third PR in a "stack" of gh-12 -> gh-16 -> this one.


ev-br:
OK, I think this is it for now, this three-PR "stack" no longer WIP.

autogen/gen_ufuncs_2.py
```diff
@@ -0,0 +1,213 @@
+from dump_namespace import grab_namespace, get_signature
```
ev-br, line 1: No need to review stuff in this file, these are all technical implements.

torch_np/tests/test_ufuncs_basic.py
```diff
@@ -131,3 +193,140 @@ def test_xy_and_out_broadcast(self, ufunc):
...
+        a = np.array([1, 2, 3])
+        b = other_dtype(3)
+
+        if ufunc in no_complex and issubclass(other_dtype, np.complexfloating):
+            pytest.skip(f'{ufunc} does not accept complex.')
+
+        # __op__
+        result = op(a, b)
+        assert_equal(result, ufunc(a, b))
+        assert result.dtype == np.result_type(a, b)
```
ev-br, line 243: OK, this (and similar assertion on `result.dtype`) is not always right. Will fix separately.

ev-br: xfail for now in the last commit. This needs fully fledged type promotion, including "weak promotion" from scalars. Will circle back to it after np.random and numpy test suite are ported.

torch_np/_ndarray.py
```diff
@@ -115,28 +119,31 @@ def __str__(self):
...
+        try:
+            return _ufunc_impl.not_equal(self, asarray(other))
+        except RuntimeError:
+            # Failed to convert other to array: definitely not equal.
+            falsy = torch.full(self.shape, fill_value=True, dtype=bool)
+            return asarray(falsy)
 
     def __gt__(self, other):
-        return asarray(self._tensor > asarray(other).get())
+        return _ufunc_impl.greater(self, asarray(other))
```
lezcano, line 137: Perhaps simpler to just call the functions in `_binary_ufuncs`? This way you could simply do
```python
import _binary_ufuncs as bfun

__gt__ = bfun.greater
...
```

lezcano: With a one-line helper function you could also golf this way all the in-place and reverse dunder methods.

torch_np/_ufunc_impl.py
```diff
@@ -0,0 +1,152 @@
...
+
+       x1, x2 are arrays! array_like -> array conversion is the caller responsibility.
+    """
+    def wrapped(x1, x2, /, out=None, *, where=True,
+                casting='same_kind', order='K', dtype=None, subok=False, **kwds):
+        _util.subok_not_ok(subok=subok)
+        if order != 'K' or not where:
+            raise NotImplementedError
+
+        # XXX: dtype=... parameter is silently ignored
```
lezcano, line 19: Let's either assert that `dtype is None` for now. But more generally, why not implementing it?

torch_np/_ufunc_impl.py
```diff
@@ -0,0 +1,152 @@
...
+def deco_unary_ufunc(torch_func):
+    # TODO: deduplicate with `deco_binary_ufunc` above. Need to figure out the
+    #       effect of the `dtype` parameter, does it differ between unary and binary ufuncs. 
+    def wrapped(x1, /, out=None, *, where=True,
+                casting='same_kind', order='K', dtype=None, subok=False, **kwds):
+        _util.subok_not_ok(subok=subok)
+        if order != 'K' or not where:
+            raise NotImplementedError
+
+        # XXX: dtype=... parameter is silently ignored
```
lezcano, line 39: as above.

torch_np/tests/numpy_tests/core/test_scalarmath.py
```diff
@@ -637,8 +638,8 @@ def test_result(self):
 class TestSubtract:
     def test_exceptions(self):
         a = np.ones((), dtype=np.bool_)[()]
-        with assert_raises(TypeError):
-            operator.sub(a, a)
+        with assert_raises((TypeError, RuntimeError)):    # XXX: TypeError from numpy
+            operator.sub(a, a)                            #  RuntimeError from torch
```
lezcano, line 642: Should this be an xfail potentially? Do we want to get in the business of raising the same exceptions at some point or is that outside of the scope?

torch_np/tests/test_basic.py
```diff
@@ -27,6 +27,8 @@
 
 
 one_arg_funcs += [getattr(w, name) for name in _unary_ufuncs.__all__]
+one_arg_funcs = one_arg_funcs[:-1]   # FIXME: remove np.invert
```
lezcano, line 30: aren't you actually removing `positive` here? If you want to remove `invert, do a `if name not "invert` in the ist comprehension above.

lezcano (approved):
+903 −1,551 I think the numbers speak for themselves!

I left a few things that I believe can be improved, but it looks mostly good!

torch_np/_ndarray.py
ev-br: Sadly, `_binary_ufuncs` has `from _ndarray import asarray`, so it's an import loop.

torch_np/_ufunc_impl.py
ev-br: Certainly will. The next batch of numpy tests will likely make me bite the bullet and figure out all the weak promotion rules, dtype resolution rules and how that interacts with registered loop types.

torch_np/tests/numpy_tests/core/test_scalarmath.py
ev-br: Would be nice to have that at some point, I guess. Finding these in the test suite would be easy though, since this is probably the only reason to catch a tuple of exceptions, so a grep stanza is straighforward.

torch_np/tests/test_basic.py
ev-br: Removing this line leads to `RuntimeError: "bitwise_not_cpu" not implemented for 'Double'`, so it certainly is np.invert.

There's a temporary technical glitch however: 

```
(Pdb) p one_arg_funcs[-1]
<function deco_unary_ufunc.<locals>.wrapped at 0x7f5a96853790>
```
Let me keep this hack as is for now (note the FIXME comment) --- until a refactor that would hopefully remove the need for two-step wrapping of ufuncs.

torch_np/_ndarray.py
lezcano: There has to be a way to untangle this. A (dirty) way of working around this for now would be to do a local import of the module within the class. But more generally, perhaps @rgommers has some ideas on how to work around these issues.

torch_np/_ufunc_impl.py
lezcano: Let's throw in an assert for now though?

torch_np/tests/numpy_tests/core/test_scalarmath.py
lezcano: sure, let's do that, but for now, let's document this somewhere so that we don't forget.

torch_np/tests/test_basic.py
lezcano: fine

torch_np/_ndarray.py
ev-br: Yes, I nearly did  a local import; however this feels almost the same level of dirty, and I'd rather hammer it until it works with a reasonable coverage, and think through a general refactor when it does. A refactor is definitely in order (together with asarray/_from_tensor_and_base rationalization we discussed in a parallel thread, a reasonable solution to asarray imports in dtypes.py etc etc)

torch_np/_ufunc_impl.py
ev-br: https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/17/commits/2b363f3a4d2e0d994608872e6e84e261e53b5d5d

torch_np/_ufunc_impl.py
ev-br: https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/17/commits/2b363f3a4d2e0d994608872e6e84e261e53b5d5d

torch_np/tests/numpy_tests/core/test_scalarmath.py
ev-br: Opened https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/20

ev-br:
Thanks for the reviews and discussions! Am merging this to make iterating on private/public split easier.
The merge commit is 
https://github.com/Quansight-Labs/numpy_pytorch_interop/commit/a8c18aac059ea8dec05809f6a71cf7658dd90b12

PR #18, opened 2023-01-09
Smoke valid args for binary ufunc tests

honno:
**This PR is WIP, and starts from #12 as it contains a critical fix.**

The binary ufunc smoke tests were passing floats to functions that didn't accept them, which this PR fixes... kinda. Running `autogen/gen_ufuncs.py` as-is does not produce the same file structure as it exists for `torch_np`, and moving things over still breaks due to invalid import references.

So I can work on fixing these things generally, although IMO it looks like these smoke tests don't need to be generated by a script, but instead could be both generated and ran on runtime with [`pytest.mark.parametrize`](https://docs.pytest.org/en/6.2.x/parametrize.html). Mind if I have a go exploring that?


ev-br:
These smoke tests certainly do not need to be generated by a script anymore, so please feel free to fix them! Also cf gh-17, which add a bunch of tests for this.

torch_np/_dtypes.py
```diff
@@ -38,14 +53,31 @@ def __eq__(self, other):
         if isinstance(other, dtype):
             return self._name == other.name
         else:
-            other_instance = dtype(other)
+            try:
+                other_instance = dtype(other)
+            except TypeError:
+                return False
```
ev-br, line 59: Not sure why it is better than raising?

torch_np/_ndarray.py
```diff
@@ -110,6 +110,16 @@ def __neq__(self, other):
...
         return asarray(self._tensor > asarray(other).get())
 
+    def __lt__(self, other):
+        return asarray(self._tensor < asarray(other).get())
+
+    def __ge__(self, other):
+        return asarray(self._tensor >= asarray(other).get())
+
+    def __le__(self, other):
+        return asarray(self._tensor <= asarray(other).get())
```
ev-br, line 120: NB: this will need to be redone similar to gh-17

torch_np/tests/test_binary_ufuncs.py
```diff
@@ -1,247 +1,232 @@
...
 import torch
 
-from .._binary_ufuncs import *
 from ..testing import assert_allclose
 
 
 def test_add():
     assert_allclose(np.add(0.5, 0.6),
-                               add(0.5, 0.6), atol=1e-7, check_dtype=False)
+                               np.add(0.5, 0.6), atol=1e-7, check_dtype=False)
```
ev-br, line 9: `assert_allclose(np.add(0.5, 0.5), np.add(0.5, 0.6))`?

honno:
Looks like the test inputs have now been fixed, so closing.

PR #21, opened 2023-01-16
Introduce `pre-commit` with `isort`/`black`/`autoflake`

honno:
Adds a [`pre-commit`](https://pre-commit.com/) config with just `black` setup, and includes modifications from a run of black (ignore-rev'd). Also includes a CI job which just checks for `pre-commit` (i.e. just `black` for now) passes.

<sup>I was wondering if folks would want other tools? i.e. personally I always like `autoflake --remove-all-unused-imports` and `isort --profile black`.</sup>


ev-br:
Re other tools: Having said  'a', I guess we can keep on all the way down to 'z'.

ev-br:
Thanks @honno !

PR #22, opened 2023-01-24
WIP: Refactor the internals to better separate wrappers from ops on tensors

ev-br:
Introduce the `_detail` subpackage: its contents only operates on pytorch tensors and does not know about wrappers. User-facing functionality mostly deals with array-likes, argument differences between numpy and pytorch, and delegates heavy lifting to functions in `_detail`.


torch_np/_detail/_reductions.py
```diff
@@ -0,0 +1,165 @@
...
+
+import torch
+from . import _util
+from . import _scalar_types
+
+NoValue = None
+
+
+def _atleast_float(dtype, other_dtype):
+    """Cast bools and ints and floats to a default float; complex to default complex.
```
rgommers, line 15: This line is a little misleading, it seems to be about scalar values rather than dtype objects. I propose:
```
Return a dtype that is real or complex floating-point.

For inputs that are boolean or integer dtypes, this returns the default float dtype;
inputs that are complex get converted to the default complex dtype;
real floating-point dtypes (`float*`) get passed through unchanged
```

And now that I've written that out - is it really the intended behavior here for `complex*` to be converted to `complex64` but `float*` to be passed through?

torch_np/_detail/_reductions.py
```diff
@@ -0,0 +1,165 @@
...
+        result = tensor.all(axis)
+    return result
+
+
+def max(tensor, axis=None, initial=NoValue,
+         where=NoValue):
+    if where is not None:
+        raise NotImplementedError
+    if initial is not None:
+        raise NotImplementedError
```
rgommers, line 79: minor: the form
```python
    if initial is not None or where is not None:
        raise NotImplementedError
```
as used further down seems nicer

rgommers: same for `min` a few liner lower

torch_np/_detail/_util.py
```diff
@@ -0,0 +1,292 @@
...
+# from https://github.com/numpy/numpy/blob/main/numpy/core/numeric.py#L1378
+def normalize_axis_tuple(axis, ndim, argname=None, allow_duplicate=False):
+    """
+    Normalizes an axis argument into a tuple of non-negative integer axes.
+    This handles shorthands such as ``1`` and converts them to ``(1,)``,
+    as well as performing the handling of negative indices covered by
+    `normalize_axis_index`.
+    By default, this forbids axes from being specified multiple times.
+    Used internally by multi-axis-checking logic.
+    .. versionadded:: 1.13.0
```
rgommers, line 54: minor: this docstring misses a bunch of blank lines

torch_np/_helpers.py
```diff
@@ -28,119 +30,38 @@ def cast_and_broadcast(arrays, out, casting):
...
-    if axis is None:
-        return axis
-    if len(axis) != 1:
-        raise NotImplementedError("does not handle tuple axis")
-    return axis[0]
-
-
-def to_tensors(*inputs):
-    """Convert all ndarrays from `inputs` to tensors."""
+def to_tensors_lax(*inputs):
```
rgommers, line 56: I saw this used elsewhere and thought it referred to [`jax.lax`](https://jax.readthedocs.io/en/latest/jax.lax.html) - it seems that's not the case, so a different name would be preferred.

rgommers (commented):
Looks pretty clean overall. A few comments from just reading the diff.

torch_np/_helpers.py
ev-br: Thanks! ` ndarrays_to_tensors` it is

torch_np/_detail/_reductions.py
ev-br: Thanks, done

torch_np/_detail/_util.py
ev-br: Restored.

torch_np/_helpers.py
ev-br: Done

torch_np/_detail/_reductions.py
ev-br: Thanks. The last commit takes over you text + makes complex dtypes pass through intact.

torch_np/_decorators.py
```diff
@@ -0,0 +1,112 @@
...
+        from ._ndarray import asarray
+        x1_tensor = asarray(x1).get()
+        result = impl_func((x1_tensor,), *args, **kwds)
+        return result
+    return wrapped
+
+
+# TODO: deduplicate with _ndarray/asarray_replacer,
+# and _wrapper/concatenate et al
+def deco_ufunc_from_impl(impl_func):
```
lezcano, line 73: Nit `deco_binary_ufunc_...`?

torch_np/_decorators.py
```diff
@@ -0,0 +1,112 @@
...
+    Note that we peel off `out=...` and `keepdims=...` args (torch functions never
+    see them). The `axis` argument we normalize and pass through to pytorch functions.
+
+    """
+    # XXX: move this out of _ndarray.py (circular imports)
+    #
+    # TODO: 1. get rid of _helpers.result_or_out
+    #       2. sort out function signatures: how they flow through all decorators etc
+    @functools.wraps(func)
+    def wrapped(a, axis=None, out=None, keepdims=NoValue, *args, **kwds):
```
lezcano, line 101: I get the feeling that this function does too much. I really liked how you split the dtype and out processing above, but here you have mixed them all together. As a result, you seem to forget handling the `out` kwarg.

torch_np/_detail/_reductions.py
```diff
@@ -0,0 +1,164 @@
...
+    return result
+
+
+def std(tensor, axis=None, dtype=None, ddof=0, *, where=NoValue):
+    if where is not None:
+        raise NotImplementedError
+
+    dtype = _atleast_float(dtype, tensor.dtype)
+
+    tensor = tensor.to(dtype)
```
lezcano, line 148: Nit.
```suggestion
    if dtype != tensor.dtype:
        tensor = tensor.to(dtype)
```
This way we avoid having `to`'s scattered all throughout the trace of the program. In PrimTorch we have an aux function that does this called `_maybe_convert_to_dtype`.

torch_np/_detail/_reductions.py
```diff
@@ -0,0 +1,164 @@
...
+    """Return a dtype that is real or complex floating-point.
+
+    For inputs that are boolean or integer dtypes, this returns the default
+    float dtype; inputs that are complex get converted to the default complex
+    dtype; real floating-point dtypes (`float*`) get passed through unchanged
+    """
+    if dtype is None:
+        dtype = other_dtype
+    if not dtype.is_floating_point:
+        if dtype.is_complex:
```
lezcano, line 24: Nit.
```suggestion
    if not dtype.is_floating_point and not dtype.is_complex:
```

torch_np/_detail/_reductions.py
```diff
@@ -0,0 +1,164 @@
...
+def _atleast_float(dtype, other_dtype):
+    """Return a dtype that is real or complex floating-point.
+
+    For inputs that are boolean or integer dtypes, this returns the default
+    float dtype; inputs that are complex get converted to the default complex
+    dtype; real floating-point dtypes (`float*`) get passed through unchanged
+    """
+    if dtype is None:
+        dtype = other_dtype
+    if not dtype.is_floating_point:
```
lezcano, line 23: This not being an `elif` means that if `x` is not a floating point tensor, then it will be casted to a floating point tensor. Is this expected?

torch_np/_detail/_reductions.py
```diff
@@ -0,0 +1,164 @@
...
+def max(tensor, axis=None, initial=NoValue, where=NoValue):
+    if initial is not None or where is not None:
+        raise NotImplementedError
+
+    result = tensor.amax(axis)
+    return result
+
+
+def min(tensor, axis=None, initial=NoValue, where=NoValue):
+    if initial is not None or where is not None:
```
lezcano, line 87: Q. Should be testing here `is not None`  or `is not NoValue`?

Also, we should probably have one utils function for each of these common args.

torch_np/_detail/_ufunc_impl.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    def wrapped(tensors, /, out_shape_dtype=None , *, where=True,
+                casting='same_kind', order='K', dtype=None, subok=False, **kwds):
+        _util.subok_not_ok(subok=subok)
+        if order != 'K' or not where:
+            raise NotImplementedError
+
+        # XXX: dtype=... parameter
+        if dtype is not None:
+            raise NotImplementedError
```
lezcano, line 21: Now implementing this should be trivial with the current infra, right?

torch_np/_detail/_ufunc_impl.py
```diff
@@ -0,0 +1,129 @@
...
+
+        tensors = _util.cast_and_broadcast(tensors, out_shape_dtype, casting)
+
+        result = torch_func(*tensors)
+        return result
+    return wrapped
+
+
+# binary ufuncs: the list is autogenerated, cf autogen/gen_ufunc_2.py
+# And edited manually! np.equal  <--> torch.eq, not torch.equal
```
lezcano, line 31: In the future we can autogenerate the whole thing and just have a list of exceptions. But LGTM either way.

torch_np/_detail/_util.py
```diff
@@ -0,0 +1,299 @@
...
+    pass
+
+
+class UFuncTypeError(TypeError, RuntimeError):
+    pass
+
+
+# a replica of the version in ./numpy/numpy/core/src/multiarray/common.h
+def normalize_axis_index(ax, ndim, argname=None):
+    if ax < -ndim or ax >= ndim:
```
lezcano, line 38: How cool is Python
```suggestion
    if not (-ndim <= ax < ndim):
```

torch_np/_detail/_util.py
```diff
@@ -0,0 +1,299 @@
...
+    Normalizes an axis argument into a tuple of non-negative integer axes.
+
+    This handles shorthands such as ``1`` and converts them to ``(1,)``,
+    as well as performing the handling of negative indices covered by
+    `normalize_axis_index`.
+
+    By default, this forbids axes from being specified multiple times.
+    Used internally by multi-axis-checking logic.
+
+    .. versionadded:: 1.13.0
```
lezcano, line 57: Conciseness: From here on, do we *really* want to keep all the docs here, or will a rough summary / removing the args do?

torch_np/_detail/_util.py
```diff
@@ -0,0 +1,299 @@
...
+    -------
+    a tuple of torch.Tensors with dtype being the PyTorch counterpart
+    of the `target_dtype` and `target_shape`
+    """
+    if out_param is None:
+        return tensors
+
+    target_dtype, target_shape = out_param
+
+    can_cast = _scalar_types._can_cast_impl
```
lezcano, line 200: Does this mean that a better place for this function would be the file `_scalar_types.py`?

torch_np/_ndarray.py
```diff
@@ -194,129 +162,81 @@ def __len__(self):
...
-    def __floordiv__(self, other):
-        return _ufunc_impl.floor_divide(self, asarray(other))
-
-    def __rfloordiv__(self, other):
-        return _ufunc_impl.floor_divide(self, asarray(other))
+    __floordiv__ = __rfloordiv__ = _binary_ufuncs.floor_divide
 
     def __ifloordiv__(self, other):
-        return _ufunc_impl.floor_divide(self, asarray(other), out=self)
+        return _binary_ufuncs.floor_divide(self, other, out=self)
```
lezcano, line 196: Nit, these can be done with a partial (+ a wraps if you want to preserve the name)

torch_np/_unary_ufuncs.py
```diff
@@ -1,24 +1,7 @@
-import functools
-import torch
+from ._decorators import deco_unary_ufunc_from_impl
+from ._detail import _ufunc_impl
```
lezcano, line 2: I don't quite understand this file. Should you do `from ._detail._ufunc_impl import *` or so? How come you are exporing all those functions in `__all__` but they are not defined at all in this file?

lezcano (commented):
What's the semantic difference between `torch_np/_util.py` and `troch_np/_helpers.py`? Should `helpers.py` be merged into `util.py`?

Also, feel free to stop prepending underscores in all the files / folders. I don't think it should be needed, and it makes everything a bit odd to read as you don't know what's public, what's private and what's what.

All in all, I think this is certainly going in the right direction. Now the mappings are much cleaner, which is great.

torch_np/_decorators.py
```diff
@@ -0,0 +1,112 @@
...
+        torch_dtype = None
+        if dtype is not None:
+            dtype = _dtypes.dtype(dtype)
+            torch_dtype = dtype._scalar_type.torch_dtype
+        return func(*args, dtype=torch_dtype, **kwds)
+    return wrapped
+
+
+def emulate_out_arg(func):
+    """Simulate the out=... handling *for functions which do not need it*.
```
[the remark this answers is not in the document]
ev-br: Rephrased, thanks.

torch_np/_detail/_reductions.py
ev-br: I was going to the other direction based on 

```
In [2]: t = torch.ones(4)

In [3]: t.to(t.dtype) is t
Out[3]: True
```

but OK, if it's preferable to have an explicit `if dtype != ...`, will add it everywhere.

torch_np/_detail/_reductions.py
ev-br: Thanks, changed. For now I kept the NoValue alias from numpy, not sure if it's really worth it. But changed for now anyway. 
Down the line these would certainly grow their dedicated utilities. Am going to push implementing `where` (via `torch.scatter` or some such) for after the initial coverage is more complete.

torch_np/_detail/_reductions.py
ev-br: Yes, thanks.

torch_np/_detail/_reductions.py
ev-br: Exactly. The use is e.g. `tensor.mean()` for `tensor.dtype` being integer. pytorch raises a RuntimeError, but numpy does not. The current version is not 100% numpy-compatible though for edge cases (`np.mean(dtype=int)` etc). Will fix after porting the relevant part of numpy test battery.

torch_np/_detail/_ufunc_impl.py
ev-br: The list is indeed autogenerated, and modifications are minimal. IMO it's easier to do tweak it once manually than keep the list of exceptions etc, but this is easily changed as needed indeed.

torch_np/_detail/_util.py
ev-br: So far it was a direct copy-paste for simplicity. Have now removed most irrelevant details from the docstring (numpy version added etc), do you want me to remove it all? My thinking is that a docstring is better than no docstring, but am not going to argue either way.

torch_np/_detail/_util.py
ev-br: The current logic is to keep `cast_and_broadcast` and `cast_dont_broadcast` next to each other (one is used by ufuncs, the other is used by `concatenate` et al). And wanted to keep `_scalar_types` minimal. But this can go either way, do you want me to move it?

torch_np/_ndarray.py
ev-br: Not easily because it needs to bind `self`:

```diff
diff --git a/torch_np/_ndarray.py b/torch_np/_ndarray.py
index e01fd92..2b1adf4 100644
--- a/torch_np/_ndarray.py
+++ b/torch_np/_ndarray.py
@@ -163,10 +163,7 @@ class ndarray:
 
     # add, self + other
     __add__ = __radd__ = _binary_ufuncs.add
-
-    def __iadd__(self, other):
-        return _binary_ufuncs.add(self, other, out=self)
-
+    __iadd__ = functools.partialmethod(_binary_ufuncs.add, out=self)
 
     # sub, self - other
     __sub__ = __rsub__ = _binary_ufuncs.subtract
```

triggers

```
$ pytest torch_np/ -x

<snip>

    class ndarray:
torch_np/_ndarray.py:166: in ndarray
    __iadd__ = functools.partialmethod(_binary_ufuncs.add, out=self)
E   NameError: name 'self' is not defined

```

torch_np/_unary_ufuncs.py
ev-br: doing `from ._detail._ufunc_impl import *` would be messy because the names are the same. The logic is roughly:
- `torch_np/__init__.py` does `from _unary_ufuncs import *`, which brings public names, e.g. `sin`
- `_unary_ufuncs.sin` receives wrapper arrays and other args (dtypes, casting etc). It extract the tensor from the wrapper array, and delegates to `_ufunc_impl.sin`
- `_ufunc_impl.sin` receives a tensor and other args, handles dtypes, casts and broadcasting (in the future, also `where=...)`, and delegates to `torch.sin`

`__all__` here is a technical implement for smoke testing in https://github.com/Quansight-Labs/numpy_pytorch_interop/blob/refactor/torch_np/tests/test_basic.py#L29

Can rename `__all__` to something else if that's helpful?

torch_np/_decorators.py
ev-br: In fact, `out=` arg should not be here already (removed in the last commit). 
Note the usage is `emulate_out_arg(axis_keepdims(...))`: https://github.com/Quansight-Labs/numpy_pytorch_interop/blob/refactor/torch_np/_ndarray.py#L281

This decorator itself simply unwraps ndarrays and passes heavy lifting to `_util.axis_keepdims` which does just what it says on the tin: handles axis tuples and `keepdims=True`.

Not sure how to simplify it further. Would a comment help?

The fact that the usage is not very clear is I guess a direct consequence of having separate decorators for various arguments. So if it's the direction we want to go, there will be more of this I'm afraid.

torch_np/_decorators.py
ev-br: Thanks, renamed back.

torch_np/_decorators.py
```diff
@@ -0,0 +1,112 @@
...
+    With ufuncs, `out` array can typcast and broadcast ufunc arguments, hence
+    extract the shape and dtype of the tensor which backs the `out` array
+    and pass these through.
+    """
+    def wrapped(*args, out=None, **kwds):
+        from ._ndarray import ndarray
+        if out is not None:
+            if not isinstance(out, ndarray):
+                raise TypeError("Return arrays must be of ArrayType")
+            kwds.update({'out_shape_dtype': (out.get().dtype, out.get().shape)})
```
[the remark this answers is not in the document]
ev-br: OK, moved this check to `result_or_out`. One consequence is that computations are done before error checking of `out`, but it's likely OK, as it's only slightly inefficient in case of a user error.

torch_np/_decorators.py
```diff
@@ -0,0 +1,112 @@
...
+
+from . import _dtypes
+from . import _helpers
+from ._detail import _util
+
+NoValue = None
+
+
+def dtype_to_torch(func):
+##    @functools.wraps  # XXX: why
```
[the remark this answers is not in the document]
ev-br: Thanks, this was a leftover. There will be some more surgery for `__name__` and signatures; am going to punt on this for now.

ev-br:
> What's the semantic difference between torch_np/_util.py and troch_np/_helpers.py? Should helpers.py be merged into util.py?

`torch_np/_detail/_util.py` is in the  `_detail` subpackage, hence it only knows about pytorch tensors, and knows nothing about wrapper ndarrays.

`torch_np/_helpers.py` is a collection of small helpers which do know about wrapper ndarrays and handle things like extracting tensors from ndarrays before handing the work over to utilities in `_detail.py`. 

Maybe it makes sense to merge `helpers.py` and `decorators.py`, if that's wanted (I personally am on a fence).

torch_np/_decorators.py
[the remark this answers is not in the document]
lezcano: I do not understand yet the part that says `*for functions which do not need it*`. Can you elaborate?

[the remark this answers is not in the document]
ev-br: That was just poor phrasing, here's the updated one, hope it's clearer:

https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/22/files#diff-5f6cbe3af37159a9ce7670eebdd5dcd13180caeca037575535cbae05d316adaaR24

torch_np/_decorators.py
lezcano: So, there are two general issues here.
I don't see why would you always need to use this in conjunction to `emulate_out_arg`. These two functions may be used independently, so having `out` in the signature of this wrapper is unnecessary.
Even more, not only is unnecessary, but it's also incorrect. If you use it with `emulate_out_arg` afterwards, you happen to get the right behaviour almost by chance. Now, if you swap the order, you'll get a function with the same signature, but that takes an `out=` kwarg and... discards it!

As mentioned, I think that this function and the one that implements the out behaviour do two very different things, and should be independent of eachother.

torch_np/_detail/_reductions.py
lezcano: It'd be good to have a wrapper that does the cast iff the dtypes are the same. For eager, you are right, this is a noop. Now, when tracing through the program, these calls do show up, and would need to be cleaned up later on by the compiler. Guarding for this helps a bit the compiler / serialisation process in the long run.

torch_np/_detail/_reductions.py
lezcano: Libraries allowing implicit casting makes for some incredibly obscure semantics.

torch_np/_detail/_reductions.py
lezcano: We have `torch.where` and it's a particularly efficient function in PyTorch 2.0 in CUDA, so well worth checking that one.

torch_np/_detail/_ufunc_impl.py
lezcano: By autogenerate, I meant that we could do something along the lines.
```python
__all__ = ["add", "arctan2", ..., "bitwise_left_shift", ...]
non_regular = {"bitwise_left_shift": "left_shift" ...}

def register_function(name):
    numpy_name = non_regular.get(name, name)
    globals()[numpy_name] = deco_ufunc(getattr(torch, name))


for name in __all__:
    register_function(name)
```

torch_np/_detail/_util.py
lezcano: Up to you, I don't really care. What'd be really useful is to leave a pemalink to where you got it from.

torch_np/_detail/_util.py
lezcano: up to you, I'm fine with either.

torch_np/_ndarray.py
lezcano: You are completely right, nvm.

torch_np/_unary_ufuncs.py
lezcano: Note that `__all__` has a very specific meaning in Python (see e.g. https://stackoverflow.com/a/64130/5280578). It should be clear after reading that why using `__all__` here is incorrect.

What I would suggest doing is:
- Set the `__all__` variable correctly everywhere to express what is it that you mean to export publicly from that file.
- After doing that, pulling the relevant functions from a given file (in that `test_basic.py` or elsewhere) should be trivial.

lezcano:
About the `_helpers.py` and `_utils.py`: If `_helpers.py` contains decorators, I would rename it to `_decorators.py`. That would simplify things.

A point that was not discussed: Why not removing the underscore that precedes all the names of all the files in the project?

About the general structure of the project and what this PR aims to achieve: I think it's going in the right direction, but it can be further improved. The reason I say this is because there are still a number of local imports, which means that there are still some code smells.
I propose still further refactor the code as follows, to have a rather definite "chain of command" when it comes to which files do what:

- We have an `ndarray` class in `torch_np/ndarray.py`. This class is nothing but a wrapper around a `torch.Tensor`.
- Within `torch_np/ndarray.py` we have a function `wrap_ndarray` that wraps an `np.ndarray` or a list or any other weird thing NumPy accepts into an `ndarray.ndarray` (i.e. it generates from those a `torch.Tensor` and wraps it in an `ndarray.ndarray`).

`wrap_ndarray` will need some support functionality, like that in `torch_np/{dtypes,scalar_types}.py` and otherwise.

We will also need to cast other NumPy-specific arguments into PyTorch specific arguments (e.g. `np.float32` -> `torch.float32`). We will call this process of going from NumPy args to PyTorch args (also the one performed by `wrap_ndarray`) "normalisation".

We will have a number of functions that perform normalisation on dtype args, ndarray args, axis args, etc. All these functions will take one parameter and will return a normalised version of it.

Summarising what we have until now, we have that in the top level of the `torch_np/` folder we have an `ndarray` class and a number of normalisation functions. We have not discussed yet how to use these normalisation functions, we'll do that in a bit.

- In the inner folder, (we may call it `np_refs`, in reference to `_refs/` in the PrimTorch project), we will implement all the relevant mappings.
  - These will assume that the inputs are `torch.Tensor`s or normalised inputs, and they will return `torch.Tensor`s. These functions will, of course, implement the NumPy behaviour using `torch.Tensor`.
  - These functions will always assume that their inputs are normalised.
- In this folder we will also implement wrappers that help implementing generic functionality. One such a example is the `out_wrapper`, but there may be others. See for example the wrappers in https://github.com/pytorch/pytorch/blob/master/torch/_prims_common/wrappers.py.

Consider how to implement now a non-trivial example like `concatenate`. This function needs to perform some non-standard preprocessing on the inputs (and perhaps some non-standard postprocessing). Given that we cannot normalise the inputs to this function in a generic way, we will implement a specific wrapper for it in the top level folder that does the preprocessing of the inputs (does all the necessary checks and sets the relevant `dtype` correctly). Then, these normalised inputs are passed to the `concatenate` function, which would just see `torch.Tensor`s and `torch.dtype`s and will return a `torch.Tensor`. And that's it.
 
The point that we have not discussed is how to map the normalisation functions to their relevant args. I propose doing so via some light typing. Let's do a simple example where we omit plenty of kwargs.
```python
def add(a: Tensor, b: Tensor, *, dtype: Optional[torch.dtype]) -> Tensor
```
(note that the `dtype` kwarg may be implemented generically via a decorator, or even better, all those params could be implemented generically for all ufuncs.

We would then have a function on the main folder with signature:
```
GenericNdArray = Union[np.ndarray, list[float], NumpyScalar]  # Incomplete list of types. NumpyScalar will be its own union type
def add(a: GenericNdArray, b: GenericNdArray, *, dtype: Optional[np.dtype]):
```
Note: We don't need to annotate the returned type. We will see why afterwards.

Then, we would have have a number of normalisation functions that would look something like as:
```python
import np_refs

def bind(fn):
    def inner(np_interface):
        fn_params = inspect.signature(np_interface).parameters
        for normalization in NORMALIZATIONS:
            fn = normalize(torch_impl, fn, normalization)
        return fn
    return inner


def normalize(fn, fn_params, normalization, condition):
    any = False
    @wraps(fn)
    def _fn(*args, **kwargs):
        fn_params = inspect.signature(foo).parameters

        # edit inplace or should we just zip the parameters and the args create a new list?
        for i, param in enumerate(list(fn_params)[:len(args_count)]):
            if condition(arg, fn_params[param])
                any = True
                args[i] = normalization_fn(args[i])

        # same as above
        for kwarg, value in kwargs.items():
            v = kwargs[kwarg]
            if condition(v, fn_params[kwarg])
                any = True
                kwargs[kwarg] = normalization_fn(v)
        return fn(*args, **kwargs)

    # optimisation to avoid wrappng things unnecessarily
    return _fn if any else fn


def normalize_ndarray(fn):
    """ Wrap parameters that accept any array-like object """
    return normalize(fn, wrap_ndarray, lambda val, _type: _type == GenericNdArray)

def normalize_tensor(fn):
    """ Wrap parameters that accept ndarrays """
    return normalize(fn, torch.from_numpy, lambda val, _type: _type == torch.Tensor)

def normalize_dtype(fn):
    ...

def normalize_output(fn):
    # TODO implement wrapping optimisation
    @wraps(fn)
    def _fn(*args, **kwargs):
        out = fn(*args, **kwargs)
        # TODO implement support for tuples
        return ndarray(out) if isinstance(out, torch.Tensor) else out

    return _fn

NORMALIZATIONS = [v.__name__.startswith("normalize_") for v in globals()]

@bind(np_refs.add)
def add(a: GenericNdArray, b: GenericNdArray, *, dtype: Optional[np.dtype]):
    pass
```

You could even do all this within the `ndarray.py` file, and then afterwards create in a forloop all the `ndarray` methods from these.

rgommers:
> A point that was not discussed: Why not removing the underscore that precedes all the names of all the files in the project?

If there's now an underscore in `_detail` then the files in that subdir don't need an underscore. In general it's very good practice to add underscores to filenames that are not public modules. Most projects aren't great at it, but it's the only way to prevent people doing `import somepkg.some_private_file` and it looking public.

rgommers:
The overall structure you propose seems nice and logical @lezcano.
 
> The point that we have not discussed is how to map the normalisation functions to their relevant args. I propose doing so via some light typing. 

This I'm not 100% sure about. Typing for the PyTorch layer is straightforward; the `GenericNdArray` part is not. It's not completely clear that an incomplete set of types in that union will be sufficient. And it's very hard to be complete here - have a look at how NumPy implements its `NDArray` static type will show that quickly enough.

lezcano:
>It's not completely clear that an incomplete set of types in that union will be sufficient. And it's very hard to be complete here

My point here is that this will serve as some annotations to:
- Have the tools to apply the normalisation functions via introspection
- Have a place where we sort of annotate (a subset of) the types the normalisation functions support. The set of supported types may be larger just by chance, but that's alright.

Note that these annotations are not meant to be used to strictly typecheck programs.

torch_np/_unary_ufuncs.py
ev-br: Well, one other meaning of  `__all__` is "these names are meant to be public even though star imports are bad idea and we do not recommend them". Still other use is to do a star import from a private submodule in `__init__.py`  --- which is exactly the thing here. 

I am going to push  setting `__all__` correctly to a next refactoring pass.

lezcano: >  Still other use is to do a star import from a private submodule in `__init__.py` --- which is exactly the thing here.

This relates to why this felt weird to me in the first place. You are setting the `__all__` attribute without making those functions available from that file, which is odd. We can discuss it in tomorrow's meeting, and I may very well be in the wrong here, but I've certainly never seen this pattern before.

torch_np/_detail/_ufunc_impl.py
ev-br: Opened https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/31 to keep track of this. FWIW, I don't know if doing it repeatedly at import time is any better than doing it once, but if it helps to upstream it, then ¯\_(ツ)_/¯

lezcano: It's not about upstreaming it, but about being concise and less error-prone. Note that this does pretty much as much work as defining the functions the way you are doing now, modulo some tiny overhead from the for-loop and the function calls and so.

ev-br:
As discussed, let's merge this one as is and discuss the next restructure in  https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/32

ev-br:
OK, no, cannot merge this. I am failing to appease the powers of linting to merge this with main with black and all. Am giving up for today. Hopefully @honno would be able to give me a hand with this.

PR #25, opened 2023-01-27
add a minimal `np.random` analog

ev-br:
Only mimic the old numpy random API.
No guarantees about reproducibility etc beyond those of pytorch.

For the moment, only add a minimal subset to help porting the numpy test suite.

Is on top of gh-22, for clarity.

### DIFF
--- a/torch_np/random.py
+++ b/torch_np/random.py
@@ -1,0 +1,120 @@
+"""Wrapper to mimic (parts of) np.random API surface.
+
+NumPy has strict guarantees on reproducibility etc; here we don't give any.
+
+Q: default dtype is float64 in numpy
+
+"""
+from math import sqrt
+
+import torch
+
+from . import asarray
+from ._detail._scalar_types import default_float_type as _default_float_type
+from ._detail import _util
+
+_default_dtype = _default_float_type.torch_dtype
+
+__all__ = ["seed", "random_sample", "sample", "random", "rand", "randn", "normal"]
+
+
+def seed(seed=None):
+    if seed is not None:
+        torch.random.manual_seed()
+
+
+def random_sample(size=None):
+    if size is None:
+        values = torch.rand(())
+        return float(values)
+    else:
+        values = torch.rand(size).to(_default_dtype)
+        return asarray(values)
+
+
+def rand(*size):
+    return random_sample(size)
+
+
+sample = random_sample
+random = random_sample
+
+
+def uniform(low=0.0, high=1.0, size=None):
+    if size is None:
+        values = torch.rand(())
+        return float(low + (high - low) * values)
+    else:
+        values = torch.rand(size).to(_default_dtype)
+        return asarray(low + (high - low) * values)
+
+
+def randn(*size):
+    if size == ():
+        return float(torch.randn(size))
+    else:
+        values = torch.randn(*size).to(_default_dtypes)
+        return asarray(values)
+
+
+def normal(loc=0.0, scale=1.0, size=None):
+    if size is None:
+        size = ()
+    return loc + scale * randn(*size).to(_default_dtype)
+
+
+def shuffle(x):
+    x_tensor = asarray(x).get()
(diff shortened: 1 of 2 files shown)


torch_np/random.py
```diff
@@ -0,0 +1,120 @@
...
+
+__all__ = ["seed", "random_sample", "sample", "random", "rand", "randn", "normal"]
+
+
+def seed(seed=None):
+    if seed is not None:
+        torch.random.manual_seed()
+
+
+def random_sample(size=None):
```
lezcano, line 26: Why not simply calling this `random` or `sample`?

torch_np/random.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+sample = random_sample
+random = random_sample
+
+
+def uniform(low=0.0, high=1.0, size=None):
+    if size is None:
+        values = torch.rand(())
+        return float(low + (high - low) * values)
```
lezcano, line 46: You may prefer to simply use `torch.uniform`.

Also, note that this formula overflows for `high = max_float, low=-max_float`. To account for this, you may want to divide high and low by 2 if they are too large and too small respectively.

torch_np/random.py
```diff
@@ -0,0 +1,120 @@
...
+
+def rand(*size):
+    return random_sample(size)
+
+
+sample = random_sample
+random = random_sample
+
+
+def uniform(low=0.0, high=1.0, size=None):
```
lezcano, line 43: You may want to write this function either using `torch.uniform` or doing `low + (high - low) * random(size)` (+ some proper handling of the overflow if you see fit).

torch_np/random.py
```diff
@@ -0,0 +1,120 @@
...
+        values = torch.rand(())
+        return float(low + (high - low) * values)
+    else:
+        values = torch.rand(size).to(_default_dtype)
+        return asarray(low + (high - low) * values)
+
+
+def randn(*size):
+    if size == ():
+        return float(torch.randn(size))
```
lezcano, line 54: Can you encapsulate all this `size` logic into its own function?

torch_np/random.py
```diff
@@ -0,0 +1,120 @@
...
+    else:
+        values = torch.randn(*size).to(_default_dtypes)
+        return asarray(values)
+
+
+def normal(loc=0.0, scale=1.0, size=None):
+    if size is None:
+        size = ()
+    return loc + scale * randn(*size).to(_default_dtype)
```
lezcano, line 63:
```suggestion
    return loc + scale * randn(*size)
```

lezcano (approved):
Just a few minor comments.

torch_np/random.py
ev-br: Why? Note that numpy's default is float64 not float32, and ISTM it's better to centralize this.

lezcano: This logic of setting the default type and handling the size is already implemented in the `randn` function, so no need to repeat it here.

torch_np/random.py
ev-br: Numpy has these three aliases (random, random_sample and sample), all three are here.

lezcano: ah, fair enough!

torch_np/random.py
ev-br: OK, refactored this to use `torch.empty(size).uniform_(...)`.
Am not going to even wonder why  `torch.rand` but  `torch.Tensor.uniform_` and no `torch.uniform`.

torch_np/random.py
ev-br: Ah, OK.  This is different in the last commit now, which implements 
`torch.empty(size, dtype=_default_dtype).uniform_(low, high)`, as requested.

torch_np/random.py
lezcano: large library, organically grown... jajajaj

PR #30, opened 2023-01-30
make ndarray.strides byte strides to match numpy

ev-br:


ev-br:
A trivial fixup, merging.

PR #36, opened 2023-02-02
BUG: testing: fix repr error on failure

ev-br:
So that the error report shows the number of failures, mismatched elements etc.


ev-br:
A trivial fix, merging.

PR #37, opened 2023-02-03
BUG: random: randint returns integers, not floats

ev-br:
`random.randint(3)` should return an integer, not a float.


ev-br:
Straightforward fix, merging.

PR #39, opened 2023-02-04
implement flip-based array manipulations

ev-br:
- [x] flip
- [x] flipup
- [x] fliplr
- [x] moveaxis
- [x] rollaxis


torch_np/_detail/_flips.py
```diff
@@ -0,0 +1,52 @@
...
+    if start < 0:
+        start += n
+    msg = "'%s' arg requires %d <= %s < %d, but %d was passed in"
+    if not (0 <= start < n + 1):
+        raise _util.AxisError(msg % ("start", -n, "start", n + 1, start))
+    if axis < start:
+        # it's been removed
+        start -= 1
+    if axis == start:
+        return tensor[...]
```
lezcano, line 48: Is this missing or what is this supposed to mean? I don't think this is a construction in PyTorch.

torch_np/_detail/_flips.py
```diff
@@ -0,0 +1,52 @@
...
+
+# https://github.com/numpy/numpy/blob/v1.24.0/numpy/core/numeric.py#L1259
+def rollaxis(tensor, axis, start=0):
+    n = tensor.ndim
+    axis = _util.normalize_axis_index(axis, n)
+    if start < 0:
+        start += n
+    msg = "'%s' arg requires %d <= %s < %d, but %d was passed in"
+    if not (0 <= start < n + 1):
+        raise _util.AxisError(msg % ("start", -n, "start", n + 1, start))
```
lezcano, line 43: Two small things.
- It'd be good to upgrade these to f-strings
- It'd be good to have a `_util.check` utility. See for example https://github.com/pytorch/pytorch/blob/8c26ed5f5e36bf0692b381b16fe1b9efdba16185/torch/_prims_common/__init__.py#L1556-L1566

torch_np/tests/numpy_tests/core/test_numeric.py
```diff
@@ -1251,78 +1251,60 @@ def test_failed_itemsetting(self):
...
+    def test_count_nonzero_list(self):
+        lst = [[0, 1, 2, 3], [1, 0, 0, 6]]
+        assert np.count_nonzero(lst) == 5
+        assert_array_equal(np.count_nonzero(lst, axis=0), np.array([1, 1, 1, 2]))
+        assert_array_equal(np.count_nonzero(lst, axis=1), np.array([3, 2]))
+
     def test_nonzero_trivial(self):
         assert_equal(np.count_nonzero(np.array([])), 0)
-        assert_equal(np.count_nonzero(np.array([], dtype='?')), 0)
+        assert_equal(np.count_nonzero(np.array([], dtype="?")), 0)
```
lezcano, line 1263: If this comes from black, it'd be good to separate blackening of files into their own PRs to avoid longer diffs.

torch_np/tests/numpy_tests/core/test_numeric.py
```diff
@@ -3002,12 +2793,12 @@ def test_nd(self, nd):
 
         # only one
         x[...] = False
-        x.flat[0] = True
+        x.ravel()[0] = True
```
lezcano, line 2796: What's the reason for this change?

torch_np/tests/numpy_tests/lib/test_function_base.py
```diff
@@ -130,6 +130,7 @@ def test_axes(self):
         assert_raises(np.AxisError, np.flip, np.ones((4, 4)), axis=-3)
         assert_raises(np.AxisError, np.flip, np.ones((4, 4)), axis=(0, 3))
 
+    @pytest.mark.xfail(reason='no [::-1] indexing')
```
lezcano, line 133: More than fair. You can even skip these I'd say, as there's no hope PyTorch will support this I don't think.

lezcano (approved):
All the comments above are rather minor really.

torch_np/_detail/_flips.py
ev-br: In numpy indexing with ellipsis creates a view (numpy view, not pytorch view):
```
In [62]: np.shares_memory(a[...], a)
Out[62]: True
```

Same in Pytorch, really:

```
In [70]: t[...]._base is t
Out[70]: True
```

What is the relevant pytorch idiom then?

torch_np/_detail/_flips.py
ev-br: Why do you people do not like the old-style string formatting :-). Nevermind :-).

On a more serious note, this whole function is a weird one out, retained in numpy for backwards compatibility:
https://stackoverflow.com/questions/29891583/reason-why-numpy-rollaxis-is-so-confusing

I'd much rather keep it as is with a permalink to the source and a link to this SO answer by the author of the improved numpy function.

Why implementing it at all. Well, maintained numpy consumers have predominantly moved to `moveaxis`. IIRC when we started this project, there was a talk about less well maintained user bases.
So I'd really prefer to just clone it from numpy with minimal mods and forget :-).

torch_np/tests/numpy_tests/core/test_numeric.py
ev-br: Yes indeed. I'm going to refrain from commenting about black et al, at all.

torch_np/tests/numpy_tests/core/test_numeric.py
ev-br: Shrug.
A technical reason is that ravel is implemented and flat is not. 
For this test specifically, there is no difference, the test passes in both pristine numpy and our `torch_np`.

Going forward, I have a feeling that `ravel()` is a common thing to do, while `.flat` is mainly from very old code, so I'd rather push it into "yeah, implement some day" bucket.

torch_np/_detail/_flips.py
lezcano: What about simply returning `tensor`. I know it's less precise, but views are handled differently in PyTorch autograd, and this would be a somewhat unnecessary pemisation.

Otherwise, the more PyTorch-y way of creating a view of a tensor would be  `t.view(t.shape)`. But again, perhaps returning `tensor` here is good enough.

torch_np/_detail/_flips.py
lezcano: sure, let's leave it.

torch_np/tests/numpy_tests/core/test_numeric.py
lezcano: cool.

torch_np/_detail/_flips.py
ev-br: OK, let's do that, thanks. Have added a code comment, too.

PR #44, opened 2023-02-06
Fix  __rop__ dunders

ev-br:
Also restore np.zeros defaulting to float64


ev-br:
A fixup to unblock main. Merging.

PR #46, opened 2023-02-07
WIP: implement {array_, VHD}split

ev-br:
The main difference with pytorch is that numpy expects `indices_or_sections` as indices into the array to split, while pytorch takes lengths of sections. Also some differences in what exactly happens when the length of the array to split is not equally divisible into sections. And some edge cases for when there are empty sections.

### DIFF
--- a/torch_np/_detail/implementations.py
+++ b/torch_np/_detail/implementations.py
@@ -46,7 +46,7 @@ def split_helper_int(tensor, indices_or_sections, axis, strict=False):
         num, sz = l % n, l // n + 1
         lst = [sz] * num
 
-   lst += [sz - 1] * (n - num)
+    lst += [sz - 1] * (n - num)
 
     result = torch.split(tensor, lst, axis)
 
@@ -61,7 +61,7 @@ def split_helper_list(tensor, indices_or_sections, axis, strict=False):
     lst = [x for x in indices_or_sections if x <= tensor.shape[axis]]
     num_extra = len(indices_or_sections) - len(lst)
 
-    lst = lst + [tensor.shape[axis]]
+    lst.append(tensor.shape[axis])
     lst = [
         lst[0],
     ] + [a - b for a, b in zip(lst[1:], lst[:-1])]
--- a/torch_np/_detail/implementations.py
+++ b/torch_np/_detail/implementations.py
@@ -46,14 +46,7 @@ def split_helper_int(tensor, indices_or_sections, axis, strict=False):
         num, sz = l % n, l // n + 1
         lst = [sz] * num
 
-        if n > l:
-            lst += [0] * (n - l)
-        else:
-            lrest = l - num * sz
-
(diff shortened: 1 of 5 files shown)


ev-br:
Also `tile` and `kron`, let's pick low-hanging fruit now that I'm looking at this test file.

torch_np/_detail/implementations.py
```diff
@@ -15,3 +15,63 @@ def tensor_equal(a1_t, a2_t, equal_nan=False):
...
+
+def split_helper_list(tensor, indices_or_sections, axis, strict=False):
+    if not isinstance(indices_or_sections, list):
+        raise NotImplementedError("split: indices_or_sections: list")
+    # numpy expectes indices, while torch expects lengths of sections
+    # also, numpy appends zero-size arrays for indices above the shape[axis]
+    lst = [x for x in indices_or_sections if x <= tensor.shape[axis]]
+    num_extra = len(indices_or_sections) - len(lst)
+
+    lst = lst + [tensor.shape[axis]]
```
lezcano, line 71: Nit. prefer append to avoid creating a new list.

torch_np/_wrapper.py
```diff
@@ -158,6 +159,77 @@ def stack(arrays, axis=0, out=None, *, dtype=None, casting="same_kind"):
...
 
 
+def array_split(ary, indices_or_sections, axis=0):
+    tensor = asarray(ary).get()
+    base = ary if isinstance(ary, ndarray) else None
+    axis = _util.normalize_axis_index(axis, tensor.ndim)
+
+    result = _impl.split_helper(tensor, indices_or_sections, axis)
+
+    return tuple(maybe_set_base(_, base) for _ in result)
```
lezcano, line 169: `_` is used for variables that are not used, but here it's being used. Rename it to `x`

torch_np/_wrapper.py
```diff
@@ -158,6 +159,77 @@ def stack(arrays, axis=0, out=None, *, dtype=None, casting="same_kind"):
...
+
+
+def split(ary, indices_or_sections, axis=0):
+    tensor = asarray(ary).get()
+    base = ary if isinstance(ary, ndarray) else None
```
lezcano, line 174: In general, I think that `base` should be dropped. Too big a hassle for an almost-negligible win. It also makes the wrapper heavier, which is not good.

lezcano (approved):
Cool, just a few minor points.

torch_np/_wrapper.py
ev-br: Let me open https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/47 and push the decision to a latter stage.

ev-br:
Thanks Mario for the review!

torch_np/_wrapper.py
lezcano: for sure

PR #49, opened 2023-02-09
Basic `ndarray.flags` implementation

honno:
Implements [`x.flags`](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flags.html), which mostly raises `NotImplementedError()`. Seemed nice if just for porting NumPy's own tests.

### DIFF
--- a/torch_np/_ndarray.py
+++ b/torch_np/_ndarray.py
@@ -14,6 +14,55 @@ from ._detail import _flips, _reductions, _util
 
 newaxis = None
 
+FLAGS = [
+    "C_CONTIGUOUS",
+    "F_CONTIGUOUS",
+    "OWNDATA",
+    "WRITEABLE",
+    "ALIGNED",
+    "WRITEBACKIFCOPY",
+    "FNC",
+    "FORC",
+    "BEHAVED",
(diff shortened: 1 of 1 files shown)


torch_np/_ndarray.py
```diff
@@ -58,6 +107,17 @@ def strides(self):
...
         return self._base
 
+    @property
+    def flags(self):
+        # Note contiguous in torch is assumed C-style
+        flag_to_value = {"C_CONTIGUOUS": self._tensor.is_contiguous()}
+        if flag_to_value["C_CONTIGUOUS"]:
+            # There's no proper way to determine if a tensor is Fortran-style
+            # contiguous in torch, but at least we know it isn't when it is
+            # C-style.
```
ev-br, line 117: Strictly speaking, 1D arrays can be both C and F contiguous. Not sure how deep into that rabbit hole we need fall though.

honno: Oh right—this was a bit magical to begin with so removed.

ev-br:
LET'S, so let's get this in to keep the ball rolling. Thanks @honno

PR #50, opened 2023-02-09
Infer dtype from `fill_value` in `full()`

honno:
Something we'd want to implement anyway, and I found useful for importing some of NumPy's tests.

### DIFF
--- a/torch_np/_wrapper.py
+++ b/torch_np/_wrapper.py
@@ -311,16 +311,7 @@ def full(shape, fill_value, dtype=None, order="C", *, like=None):
     if isinstance(fill_value, ndarray):
         fill_value = fill_value.get()
     if dtype is None:
-        if isinstance(fill_value, bool):
-            torch_dtype = torch.bool
-        elif isinstance(fill_value, int):
-            torch_dtype = torch.int64
-        elif isinstance(fill_value, float):
-            torch_dtype = torch.float64
-        elif isinstance(fill_value, complex):
-            torch_dtype = torch.complex128
-        else:
-            torch_dtype = _dtypes.torch_dtype_from(dtype)
+        torch_dtype = _dtypes.torch_dtype_from(type(fill_value))
     else:
         torch_dtype = _dtypes.torch_dtype_from(dtype)
     return asarray(torch.full(shape, fill_value, dtype=torch_dtype))
--- a/torch_np/_wrapper.py
+++ b/torch_np/_wrapper.py
@@ -311,7 +311,7 @@ def full(shape, fill_value, dtype=None, order="C", *, like=None):
     if isinstance(fill_value, ndarray):
         fill_value = fill_value.get()
     if dtype is None:
-        torch_dtype = _dtypes.torch_dtype_from(type(fill_value))
+        torch_dtype = asarray(fill_value).get().dtype
     else:
         torch_dtype = _dtypes.torch_dtype_from(dtype)
     return asarray(torch.full(shape, fill_value, dtype=torch_dtype))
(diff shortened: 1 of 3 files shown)


torch_np/_wrapper.py
```diff
@@ -310,7 +310,19 @@ def full(shape, fill_value, dtype=None, order="C", *, like=None):
...
+        elif isinstance(fill_value, int):
+            torch_dtype = torch.int64
+        elif isinstance(fill_value, float):
+            torch_dtype = torch.float64
+        elif isinstance(fill_value, complex):
+            torch_dtype = torch.complex128
+        else:
+            torch_dtype = _dtypes.torch_dtype_from(dtype)
+    else:
+        torch_dtype = _dtypes.torch_dtype_from(dtype)
```
ev-br, line 325: NB: I thought the if-elif chain is taken care of already in `torch_dtype_from`:

```
In [3]: tnp._dtypes.torch_dtype_from(complex)
Out[3]: torch.complex128

In [4]: tnp._dtypes.torch_dtype_from(float)
Out[4]: torch.float64

In [5]: tnp._dtypes.torch_dtype_from(bool)
Out[5]: torch.bool

In [6]: tnp._dtypes.torch_dtype_from(int)
Out[6]: torch.int64
```
(am going to move `_dtype_from_torch` wholesale into `@_decorators.dtype_to_torch` decorator, cf `eye`).
That said, this all is a bit in flux, so as long as it's tested, this particular code can be as ugly as needed --- I'm going to take this along in the next refactor pass.

honno: `torch_dtype_from(type(fill_value))` wouldn't follow the expected behaviour for dtype inference, e.g.

```python
>>> torch_dtype_from(int)
uint8  # we need int64 for dtype inference in full()
```

honno: Yep you're right, I didn't bother to check this :facepalm: Pushed a commit that just does `torch_dtype_from(type(fill_value))`

torch_np/_wrapper.py
```diff
@@ -311,16 +311,7 @@ def full(shape, fill_value, dtype=None, order="C", *, like=None):
...
-            torch_dtype = torch.bool
-        elif isinstance(fill_value, int):
-            torch_dtype = torch.int64
-        elif isinstance(fill_value, float):
-            torch_dtype = torch.float64
-        elif isinstance(fill_value, complex):
-            torch_dtype = torch.complex128
-        else:
-            torch_dtype = _dtypes.torch_dtype_from(dtype)
+        torch_dtype = _dtypes.torch_dtype_from(type(fill_value))
```
ev-br, line 314: Sadly, this won't work because 'scalar' here can actually a zero-dim torch tensor (e.g. in reductions -- cf CI failures) and 

```
(Pdb) p type(arg)
<class 'torch._C._TensorMeta'>
(Pdb) torch.is_tensor(arg)
```

I'd suggest to roll back the last commit: I'm going to work through dtype inference anyway, so as long as the behavior is correct, I'll tidy it up. Unless you want to take a detour and tidy up the dtype handling yourself of course :-).

The usage of `np.full` in reductions is arguably a bug itself, so maybe it should be fixed in there. At any rate, this change is great that it shows that the current infra is too brittle and shows its breaking points. Which is exactly what's needed, getting data on actual usage.

honno: Ah I forgot you can pass 0d arrays as the `fill_value`, nice catch—I implemented your fix and updated the introduced `test_full` test to cover this.

ev-br:
CI is green now, the change is a definite improvement, merging. Will make further improvements in a follow-up.

Issue #32, opened 2023-01-30
Split NumPy -> PyTorch preprocessing and PyTorch implementation


ev-br:
Here's Mario's detailed suggestion: https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/22#issuecomment-1408361050
Let's discuss it in this issue instead of gh-22.

ev-br:
Done in https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/82 and gh-70

PR #51, opened 2023-02-09
Make ones/zeros/empty/full dtype handling more uniform

ev-br:
while at it, un-xfail test_numeric.py::TestTypes.

### DIFF
--- a/torch_np/_wrapper.py
+++ b/torch_np/_wrapper.py
@@ -288,12 +288,20 @@ def arange(start=None, stop=None, step=1, dtype=None, *, like=None):
         raise ValueError("Maximum allowed size exceeded")
 
 
+@_decorators.dtype_to_torch
 def empty(shape, dtype=float, order="C", *, like=None):
     _util.subok_not_ok(like)
     if order != "C":
         raise NotImplementedError
-    torch_dtype = _dtypes.torch_dtype_from(dtype)
-    return asarray(torch.empty(shape, dtype=torch_dtype))
+
+    if dtype is None:
+        from ._detail._scalar_types import default_float_type
+
+        dtype = default_float_type.torch_dtype
+
+    result = torch.empty(shape, dtype=dtype)
+
+    return asarray(result)
 
 
 # NB: *_like function deliberately deviate from numpy: it has subok=True
@@ -309,15 +317,22 @@ def empty_like(prototype, dtype=None, order="K", subok=False, shape=None):
         result = result.reshape(shape)
     return result
 
+
 @_decorators.dtype_to_torch
 def full(shape, fill_value, dtype=None, order="C", *, like=None):
     _util.subok_not_ok(like)
     if order != "C":
         raise NotImplementedError
+
     fill_value = asarray(fill_value).get()
     if dtype is None:
-        dtype =  fill_value.dtype
+        dtype = fill_value.dtype
+
+    if not isinstance(shape, (tuple, list)):
+        shape = (shape,)
+
     result = torch.full(shape, fill_value, dtype=dtype)
+
     return asarray(result)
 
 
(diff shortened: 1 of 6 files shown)


ev-br:
@honno what to you think this further tweak to `full`?

honno (approved):
Generally everything looks good. 

> what to you think this further tweak to `full`?

LGTM! Updated my exploratory test for it, didn't find any issues. Everything else I suppose we expect to get covered anyway when porting tests, if they're not already.

ev-br:
https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/51/commits/1b41e1433813d6c6243a83f44041657651153b91 implements `dtype.kind` symbol following numpy : it's basically a first character of a dtype category ('f' for all floats, 'c' for complex variants, 'b' for bool , 'i' for integers)

honno:
Niche behaviour I found in NumPy

```py
>>> np.zeros((1, 5), dtype="c")  # character dtype special case
Out[13]: array([[b'', b'', b'', b'', b'']], dtype='|S1')
```

Doubt we care to implement this, just wondering if it would be nice to check for this and raise `NotImplementedError`, i.e. right now this PR gives

```py
TypeError: data type 'c' not understood
>>>

**Comment by ev-br** [comment]
ISTM this is the best we can do with non-numeric dtypes, which are not supported by pytorch. Do you think NotImplementedError is more descriptive than TypeError?

**Comment by lezcano** [comment]
IMO `TypeError` here's good, as we don't plan to implement these.

File: torch_np/_wrapper.py [hunk 1]
```diff
@@ -288,12 +288,15 @@ def arange(start=None, stop=None, step=1, dtype=None, *, like=None):
         raise ValueError("Maximum allowed size exceeded")
 
 
+@_decorators.dtype_to_torch
 def empty(shape, dtype=float, order="C", *, like=None):
     _util.subok_not_ok(like)
```
lezcano, line 293: There's a bit of a mismatch here?

torch_np/_wrapper.py
```diff
@@ -310,17 +313,22 @@ def empty_like(prototype, dtype=None, order="K", subok=False, shape=None):
...
-    if isinstance(fill_value, ndarray):
-        fill_value = fill_value.get()
+
+    fill_value = asarray(fill_value).get()
     if dtype is None:
-        torch_dtype = asarray(fill_value).get().dtype
-    else:
-        torch_dtype = _dtypes.torch_dtype_from(dtype)
-    return asarray(torch.full(shape, fill_value, dtype=torch_dtype))
+        dtype = fill_value.dtype
```
lezcano, line 324: A better way to implement this would be having a function that given a tensor or a number returns its torch dtype. The reason here is that we don't want to wrap scalars into tensors if we don't need to. Scalars are easier to treat in a compiler than tensors, and often get extra optimisations (no memory allocations, no synchronisations when accessed if we are working on CUDA, you can specialise on them...)

Note that this applies in general, so we should be mindful about this going all across.

lezcano (approved):
There's a comment that's important all across the codebase, but otherwise the PR LGTM.

Feel free to fix that point in a future PR if you prefer.

torch_np/_wrapper.py
ev-br: What mismatch?

lezcano: shouldn't we have a function that's  `_util.like_not_implemented(like)` or so? Why are we passing `like` to `subok`?

ev-br: Ah, so it's the util name which is confusing. Conceptually, I'd say that `like` and `subok` are similar: one is specifically for ndarray subclasses, the other one is for a newer `__array_wrap__` protocol (I think). Most functions which work for subclasses or the protocol, have both arguments, some have one or the other. We are not going to support either of them. The utility is literally two lines:
https://github.com/Quansight-Labs/numpy_pytorch_interop/blob/main/torch_np/_detail/_util.py#L22

So if we go with utility functions I'd keep it as is, can rename it if really wanted. If we transform to the normalization approach of gh-32, these two would likely get their own type annotations. 

So overall I'd rather not touch it now.

torch_np/_wrapper.py
ev-br: Let's move this to https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/56, to not conflate a small cleanup of this PR with a rather large rework.

ev-br:
Merging to keep the ball rolling. Comments ask for a codebase wide tweaks, so can be addressed in a sequel.

PR #52, opened 2023-02-11
make better split between wrappers and torch implementations

ev-br:
Working towards moving actual logic to `_detail`, so that top level modules (currently, _wrappers.py, _ndarray.py) unwrap/normalize their inputs and pass heavy lifting to `_detail`. Details of organization of the `_detail` submodules are TBD, and so are details of unwrapping/normalization or arguments. This PR is to only to make the split as clean as reasonable.

### DIFF
--- a/torch_np/_detail/implementations.py
+++ b/torch_np/_detail/implementations.py
@@ -80,6 +80,17 @@ def tensor_angle(z, deg=False):
     return result
 
 
+# ### sorting ###
+
+def tensor_argsort(tensor, axis=-1, kind=None, order=None):
+    if order is not None:
+        raise NotImplementedError
+    stable = True if kind == "stable" else False
+    if axis is None:
+        axis = -1
+    return torch.argsort(tensor, stable=stable, dim=axis, descending=False)
+
+
 # ### splits ###
 
 
--- a/torch_np/_detail/implementations.py
+++ b/torch_np/_detail/implementations.py
@@ -82,6 +82,7 @@ def tensor_angle(z, deg=False):
 
 # ### sorting ###
 
+
 def tensor_argsort(tensor, axis=-1, kind=None, order=None):
     if order is not None:
         raise NotImplementedError
@@ -89,6 +90,45 @@ def tensor_argsort(tensor, axis=-1, kind=None, order=None):
     if axis is None:
         axis = -1
     return torch.argsort(tensor, stable=stable, dim=axis, descending=False)
+
+
+# ### tri*-something ###
+
+
+def tri(N, M, k, dtype):
+    if M is None:
+        M = N
+    tensor = torch.ones((N, M), dtype=dtype)
+    tensor = torch.tril(tensor, diagonal=k)
+    return tensor
+
+
+def triu_indices_from(tensor, k):
+    if tensor.ndim != 2:
+        raise ValueError("input array must be 2-d")
+    result = torch.triu_indices(tensor.shape[0], tensor.shape[1], offset=k)
+    return result
+
+
+def tril_indices_from(tensor, k=0):
+    if tensor.ndim != 2:
+        raise ValueError("input array must be 2-d")
+    result = torch.tril_indices(tensor.shape[0], tensor.shape[1], offset=k)
+    return result
+
+
+def tril_indices(n, k=0, m=None):
+    if m is None:
+        m = n
+    result = torch.tril_indices(n, m, offset=k)
+    return result
+
+
+def triu_indices(n, k=0, m=None):
+    if m is None:
+        m = n
+    result = torch.triu_indices(n, m, offset=k)
+    return result
 
 
 # ### splits ###
--- a/torch_np/_detail/implementations.py
+++ b/torch_np/_detail/implementations.py
@@ -1,6 +1,8 @@
 import torch
 
-from . import _util
+from . import _dtypes_impl, _util
+
+# ### equality, equivalence, allclose ###
 
 
 def tensor_equal(a1_t, a2_t, equal_nan=False):
@@ -17,6 +19,27 @@ def tensor_equal(a1_t, a2_t, equal_nan=False):
     else:
         result = a1_t == a2_t
     return bool(result.all())
+
+
+def tensor_equiv(a1_t, a2_t):
+    # *almost* the same as tensor_equal: _equiv tries to broadcast, _equal does not
+    try:
+        a1_t, a2_t = torch.broadcast_tensors(a1_t, a2_t)
+    except RuntimeError:
+        # failed to broadcast => not equivalent
+        return False
+    return tensor_equal(a1_t, a2_t)
+
+
+def tensor_isclose(a, b, rtol=1.0e-5, atol=1.0e-8, equal_nan=False):
+    dtype = _dtypes_impl.result_type_impl((a.dtype, b.dtype))
+    a = a.to(dtype)
+    b = b.to(dtype)
+    result = torch.isclose(a, b, rtol=rtol, atol=atol, equal_nan=equal_nan)
+    return result
+
+
+# ### splits ###
 
 
 def split_helper(tensor, indices_or_sections, axis, strict=False):
--- a/torch_np/_detail/implementations.py
+++ b/torch_np/_detail/implementations.py
@@ -72,6 +72,20 @@ def split_helper_list(tensor, indices_or_sections, axis, strict=False):
     return torch.split(tensor, lst, axis)
 
 
+def clip(tensor, t_min, t_max):
+    if t_min is not None:
+        t_min = torch.broadcast_to(t_min, tensor.shape)
+
+    if t_max is not None:
+        t_max = torch.broadcast_to(t_max, tensor.shape)
+
+    if t_min is None and t_max is None:
+        raise ValueError("One of max or min must be given")
+
+    result = tensor.clamp(t_min, t_max)
+    return result
+
+
 def diff(a_tensor, n=1, axis=-1, prepend_tensor=None, append_tensor=None):
     axis = _util.normalize_axis_index(axis, a_tensor.ndim)
 
--- a/torch_np/_detail/implementations.py
+++ b/torch_np/_detail/implementations.py
@@ -229,73 +229,66 @@ def diff(a_tensor, n=1, axis=-1, prepend_tensor=None, append_tensor=None):
 # #### concatenate and relatives
 
 
-def concatenate(tensors, axis=0, out=None, dtype=None, casting="same_kind"):
-    # np.concatenate ravels if axis=None
-    tensors, axis = _util.axis_none_ravel(*tensors, axis=axis)
+def _concat_cast_helper(tensors, out=None, dtype=None, casting="same_kind"):
+    """Figure out dtypes, cast if necessary."""
 
     if out is not None or dtype is not None:
         # figure out the type of the inputs and outputs
         out_dtype = out.dtype.torch_dtype if dtype is None else dtype
+    else:
+        out_dtype = _dtypes_impl.result_type_impl([t.dtype for t in tensors])
 
-        # cast input arrays if necessary; do not broadcast them agains `out`
-        tensors = _util.cast_dont_broadcast(tensors, out_dtype, casting)
+    # cast input arrays if necessary; do not broadcast them agains `out`
+    tensors = _util.cast_dont_broadcast(tensors, out_dtype, casting)
+
+    return tensors
+
+
+def concatenate(tensors, axis=0, out=None, dtype=None, casting="same_kind"):
+    # np.concatenate ravels if axis=None
+    tensors, axis = _util.axis_none_ravel(*tensors, axis=axis)
+    tensors = _concat_cast_helper(tensors, out, dtype, casting)
 
     try:
         result = torch.cat(tensors, axis)
-    except (IndexError, RuntimeError):
-        raise _util.AxisError
+    except (IndexError, RuntimeError) as e:
+        raise _util.AxisError(*e.args)
 
     return result
 
 
 def stack(tensors, axis=0, out=None, *, dtype=None, casting="same_kind"):
-    shapes = {t.shape for t in tensors}
-    if len(shapes) != 1:
-        raise ValueError("all input arrays must have the same shape")
-
+    tensors = _concat_cast_helper(tensors, dtype=dtype, casting=casting)
     result_ndim = tensors[0].ndim + 1
     axis = _util.normalize_axis_index(axis, result_ndim)
-
-    sl = (slice(None),) * axis + (None,)
-    expanded_tensors = [tensor[sl] for tensor in tensors]
-    result = concatenate(
-        expanded_tensors, axis=axis, out=out, dtype=dtype, casting=casting
-    )
-
+    try:
+        result = torch.stack(tensors, axis=axis)
+    except RuntimeError as e:
+        raise ValueError(*e.args)
     return result
 
 
-def column_stack(tensors_, *, dtype=None, casting="same_kind"):
-    tensors = []
-    for t in tensors_:
-        if t.ndim < 2:
-            t = _util._coerce_to_tensor(t, copy=False, ndmin=2).mT
-        tensors.append(t)
-
-    result = concatenate(tensors, 1, dtype=dtype, casting=casting)
+def column_stack(tensors, *, dtype=None, casting="same_kind"):
+    tensors = _concat_cast_helper(tensors, dtype=dtype, casting=casting)
+    result = torch.column_stack(tensors)
     return result
 
 
 def dstack(tensors, *, dtype=None, casting="same_kind"):
-    tensors = torch.atleast_3d(tensors)
-    result = concatenate(tensors, 2, dtype=dtype, casting=casting)
+    tensors = _concat_cast_helper(tensors, dtype=dtype, casting=casting)
+    result = torch.dstack(tensors)
     return result
 
 
 def hstack(tensors, *, dtype=None, casting="same_kind"):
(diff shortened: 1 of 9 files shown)


ev-br:
This is no longer WIP, this is large enough. Nothing exciting here, this is pure MAINT PR, which sets the ground for a future work on communications between array_like user API and pytorch implementations.

torch_np/_detail/implementations.py
```diff
@@ -19,11 +21,125 @@ def tensor_equal(a1_t, a2_t, equal_nan=False):
...
+        result = result.item()
+    return result
+
+
+def tensor_real_if_close(x, tol=100):
+    if not torch.is_complex(x):
+        return x
+    mask = torch.abs(x.imag) < tol * torch.finfo(x.dtype).eps
+    if mask.all():
+        return x.real
```
lezcano, line 68: This returns a view into the original tensor. We don't care too much about views / copies yet, but just FYI

torch_np/_detail/implementations.py
```diff
@@ -19,11 +21,125 @@ def tensor_equal(a1_t, a2_t, equal_nan=False):
...
+        return x
+
+
+# ### math functions ###
+
+
+def tensor_angle(z, deg=False):
+    result = torch.angle(z)
+    if deg:
+        result *= 180 / torch.pi
```
lezcano, line 79:
```suggestion
        result = result * (180 / torch.pi)
```
Let's try to avoid in-place operations whenever possible, as they are not differentiable in PyTorch.

torch_np/_detail/implementations.py
```diff
@@ -19,11 +21,125 @@ def tensor_equal(a1_t, a2_t, equal_nan=False):
...
+    return result
+
+
+# ### sorting ###
+
+
+def tensor_argsort(tensor, axis=-1, kind=None, order=None):
+    if order is not None:
+        raise NotImplementedError
+    stable = True if kind == "stable" else False
```
lezcano, line 89:
```suggestion
    stable = kind == "stable"
```

torch_np/_detail/implementations.py
```diff
@@ -95,6 +225,129 @@ def diff(a_tensor, n=1, axis=-1, prepend_tensor=None, append_tensor=None):
...
+        # figure out the type of the inputs and outputs
+        out_dtype = out.dtype.torch_dtype if dtype is None else dtype
+
+        # cast input arrays if necessary; do not broadcast them agains `out`
+        tensors = _util.cast_dont_broadcast(tensors, out_dtype, casting)
+
+    try:
+        result = torch.cat(tensors, axis)
+    except (IndexError, RuntimeError):
+        raise _util.AxisError
```
lezcano, line 245: What if there's a shape error?

torch_np/_detail/implementations.py
```diff
@@ -95,6 +225,129 @@ def diff(a_tensor, n=1, axis=-1, prepend_tensor=None, append_tensor=None):
...
+# #### concatenate and relatives
+
+
+def concatenate(tensors, axis=0, out=None, dtype=None, casting="same_kind"):
+    # np.concatenate ravels if axis=None
+    tensors, axis = _util.axis_none_ravel(*tensors, axis=axis)
+
+    if out is not None or dtype is not None:
+        # figure out the type of the inputs and outputs
+        out_dtype = out.dtype.torch_dtype if dtype is None else dtype
```
lezcano, line 237: an interesting pythonic idiom for this pattern, when `dtype` is an object (as in this case) would be
```python
out_dtype = dtype or out.dtype.torch_dtype
```

Some people like it, others don't. Just leaving it here as a reminder

torch_np/_detail/implementations.py
```diff
@@ -95,6 +225,129 @@ def diff(a_tensor, n=1, axis=-1, prepend_tensor=None, append_tensor=None):
...
+    shapes = {t.shape for t in tensors}
+    if len(shapes) != 1:
+        raise ValueError("all input arrays must have the same shape")
+
+    result_ndim = tensors[0].ndim + 1
+    axis = _util.normalize_axis_index(axis, result_ndim)
+
+    sl = (slice(None),) * axis + (None,)
+    expanded_tensors = [tensor[sl] for tensor in tensors]
+    result = concatenate(
```
lezcano, line 260: Why not implementing this in terms of `torch.stack`?

torch_np/_detail/implementations.py
```diff
@@ -95,6 +225,129 @@ def diff(a_tensor, n=1, axis=-1, prepend_tensor=None, append_tensor=None):
...
+        expanded_tensors, axis=axis, out=out, dtype=dtype, casting=casting
+    )
+
+    return result
+
+
+def column_stack(tensors_, *, dtype=None, casting="same_kind"):
```
lezcano, line 267: In general, I'd suggest implementing all these in terms of `torch.stack`, as it's their closest cousin, and it's less general than `torch.cat` (i.e. more opportunities for optimising it).

torch_np/_detail/implementations.py
```diff
@@ -118,3 +371,101 @@ def meshgrid(*xi_tensors, copy=True, sparse=False, indexing="xy"):
...
+    return result
+
+
+# ### linspace, geomspace, logspace and arange ###
+
+
+def geomspace(start, stop, num=50, endpoint=True, dtype=None, axis=0):
+    if axis != 0 or not endpoint:
+        raise NotImplementedError
+    tstart, tstop = torch.as_tensor([start, stop])
```
lezcano, line 390: Why do we have this line? Shouldn't these be tensors already here?

torch_np/_detail/implementations.py
```diff
@@ -118,3 +371,101 @@ def meshgrid(*xi_tensors, copy=True, sparse=False, indexing="xy"):
...
+    base = torch.pow(tstop / tstart, 1.0 / (num - 1))
+    result = torch.logspace(
+        torch.log(tstart) / torch.log(base),
+        torch.log(tstop) / torch.log(base),
```
lezcano, line 394: Let's help CSE here and compute `log_base`  just once.

lezcano (commented):
Now that the implementations are **much** cleaner, I went on and re-reviewed the implementations of the functions.

Now, this starts to look much better! The implementationos in `implementations.py` are much cleaner and easier to understand, and then the ugly and repetitive preprocessing lives in `_wrapper.py`.

torch_np/_detail/implementations.py
```diff
@@ -19,11 +21,125 @@ def tensor_equal(a1_t, a2_t, equal_nan=False):
     return bool(result.all())
 
 
+def tensor_equiv(a1_t, a2_t):
```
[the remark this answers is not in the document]
ev-br: Good point, done elsewhere --- `tensor_equal` and `tensor_equiv` are analogs of top level namespace `array_equal` and `array_equiv`; to make it more interesting `np.array_equal` is different from `np.equal` :-).

torch_np/_detail/implementations.py
ev-br: Does it?

```
In [2]: t = torch.arange(9)

In [3]: t.real._base

In [4]: t.real._base is None
Out[4]: True
```

Seems to be the same for the bare pytorch, numpy and the current wrapper package. Am I missing something?

torch_np/_detail/implementations.py
ev-br: Not sure what you mean. Numpy raises np.AxisError, with
```
In [17]: np.AxisError.__mro__
Out[17]: 
(numpy.AxisError,
 ValueError,
 IndexError,
 LookupError,
 Exception,
 BaseException,
 object)
```

here we raise an AxisError which is

```
class AxisError(ValueError, IndexError):
    pass
```

lezcano: My point is that `np.concatenate` does throw other errors as well:
```python
>>> import numpy as np
>>> x = np.ndarray([1,2,3])
>>> y = np.ndarray([1,2,3,5])
>>> np.concatenate([x,y], axis=0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<__array_function__ internals>", line 200, in concatenate
ValueError: all the input arrays must have same number of dimensions, but the array at index 0 has 3 dimension(s) and the array at index 1 has 4 dimension(s)
```
I think it's fine not to be too fine with errors just yet, as we may sometimes mask other errors, ad that's not nice.

torch_np/_detail/implementations.py
ev-br:
```
TypeError: pow() received an invalid combination of arguments - got (float, float), but expected one of:
 * (Tensor input, Tensor exponent, *, Tensor out)
 * (Number self, Tensor exponent, *, Tensor out)
 * (Tensor input, Number exponent, *, Tensor out)
```

is a one of those counter-examples of pytorch being mostly consistent :-).

Will add conversion to tensors to the wrapper to make the internt clearer then.

lezcano: PyTorch functions never accept two numbers. If this gets two numbers, why don't you use `a ** b`? You'll need to do the same thing by choosing between `torch.log` and `math.log`.

torch_np/_detail/implementations.py
lezcano: It's  not that it's a copy in the real case, but it's the tensor itself.
```python
>>> x = torch.arange(9)
>>> x.real is x
True
```

lezcano: Note that in the complex case it's a view of every other element in the tensor:
```python
>>> t = torch.arange(9).cfloat()
>>> t.real.stride()
(2,)
```

torch_np/_detail/implementations.py
ev-br: Yes. For the time being though, I think we're OK with just raising the right error class.

lezcano: My point is that sometimes it's not even the right error class. In a similar case
```python
>>> torch.cat([torch.arange(4).unsqueeze(1), torch.arange(5).unsqueeze(1)], dim=1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: Sizes of tensors must match except in dimension 1. Expected size 4 but got size 5 for tensor number 1 in the list.
```
you'd be swallowing that error and returning an incorrect `AxisError` with no message, which is super confusing.

I think the treatment of errors is best left for the future. The main reasons being:
- Simplicity of our code
- It's super difficult to be consistent here
- One of our starting goals was not to fully imitate NumPy, but to imitate **correct** NumPy programs, so that leaves errors pretty much out of scope.

torch_np/_detail/implementations.py
ev-br: Wow.
I suspect a program which critically depends on these things is an entry for a code golf competition, so I agree it's premature to seriously worry about it.

torch_np/_detail/implementations.py
```diff
@@ -19,11 +21,125 @@ def tensor_equal(a1_t, a2_t, equal_nan=False):
...
+    except RuntimeError:
+        # failed to broadcast => not equivalent
+        return False
+    return tensor_equal(a1_t, a2_t)
+
+
+def tensor_isclose(a, b, rtol=1.0e-5, atol=1.0e-8, equal_nan=False):
+    dtype = _dtypes_impl.result_type_impl((a.dtype, b.dtype))
+    a = a.to(dtype)
+    b = b.to(dtype)
```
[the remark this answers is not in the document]
ev-br: Done in https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/52/commits/13e52778f48140101146a37ecec2ef3cced77539

torch_np/_detail/implementations.py
lezcano: Hmmmm, not really. If you are interested in performance and you don't care about autograd (for example, when you have deployed a trained model and you just want to run it), one of the things you want to minimise is the number of allocated tensors throughout the program. For that, views and writing-through-views are essential. If in one place you return a view, and in other you return a copy... then that's game.

torch_np/_detail/implementations.py
ev-br: Here specifically, numpy raises a ValueError, 

```
[31]: np.concatenate((np.arange(4)[:, None], np.arange(3)[:, None]), axis=1)
ValueError                                Traceback (most recent call last)
Input In [31], in <cell line: 1>()
----> 1 np.concatenate((np.arange(4)[:, None], np.arange(3)[:, None]), axis=1)

File <__array_function__ internals>:200, in concatenate(*args, **kwargs)

ValueError: all the input array dimensions except for the concatenation axis must match exactly, but along dimension 0, the array at index 0 has size 4 and the array at index 1 has size 3
```

so a try-except catches it. 

That said, we're on the same page ISTM, on leaving fine details of having nice error messages to a yet unspecificied later point in the lifecycle of this project.

torch_np/_detail/implementations.py
ev-br: stack seems to be slower? https://stackoverflow.com/questions/54307225/whats-the-difference-between-torch-stack-and-torch-cat-functions

The way these are written at the moment is a straightforward vendor of what numpy does. So was straighforward for me at least. 
Is this SO answer still relevant?

lezcano: sorry, but where in that question says that stack is slower?

Something that's relevant here is what one of the comment says, that stack can be implemented using `cat` + views, which is the point I made above.

torch_np/_detail/implementations.py
ev-br: FTR, my background is _deeply_ caring for performance, and I'm regularly serving as a PITA reviewer with perf concerns in several other projects. And I do appreciate why views are important.

So _critically depends_ comment was about https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/52#discussion_r1110032280 : where it's a copy vs view depending on the input dtype. 

In fact, I'd argue that if the view semantics is important here, the behavior of `torch.arange(9).real` is a bug.

torch_np/_detail/implementations.py
ev-br: The answer by _casualcausality_, from March 8, 2022:

```
import torch
torch.__version__ # 1.10.2
x = [torch.randn(1) for _ in range(10000)]
torch.cat(x).shape, torch.stack(x).shape # torch.Size([10000]), torch.Size([10000, 1])

%timeit torch.cat(x) # 1.5 ms ± 476 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%timeit torch.cat(x).reshape(-1,1) # 1.95 ms ± 534 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%timeit torch.stack(x) # 5.36 ms ± 643 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

torch_np/_detail/implementations.py
lezcano: Note that `.real` is **never** a copy! It's either the original tensor or a view into the original tensor. A purist could say "oh, but you then should return the identity view onto the original tensor if the dtype is real" but it doesn't really matter. In this case, I do agree that if a program depends on being a view or the original tensor, then, well... you sort of deserve it.

Note that the PrimTorch project **does** go the extra mile and gets this one right. There is a `prims.view_of` function for this exact purpose. Again, I believe it's not necessary for us to do this.

torch_np/_detail/implementations.py
lezcano: I think that things have got better in the last few versions:
```python
In [1]: import torch
   ...: torch.__version__ # 1.10.2
   ...: x = [torch.randn(1) for _ in range(10000)]

In [2]: %timeit torch.cat(x)
918 µs ± 1.87 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [3]: %timeit torch.cat(x).reshape(-1,1)
   ...: oops each)
917 µs ± 19.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [4]: %timeit torch.stack(x)
947 µs ± 2.65 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
so I wouldn't worry about it.

torch_np/_detail/implementations.py
ev-br: OK, am rolling back gratuitous `.clone()` calls, and let's keep this in mind for later.

torch_np/_detail/implementations.py
ev-br: Okay, https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/52/commits/6970fd3efdfa0044a06431db03b7922b0c32b577 make all these *stack variants delegate to their eponymous pytorch equivalents. I guess it's the best we can do in the wrapper layer.

lezcano (approved):
Good stuff! This is certainly a great step in the right direction.

torch_np/_detail/implementations.py
```diff
@@ -118,3 +368,101 @@ def meshgrid(*xi_tensors, copy=True, sparse=False, indexing="xy"):
...
+        M = N
+    z = torch.zeros(N, M, dtype=dtype)
+    z.diagonal(k).fill_(1)
+    return z
+
+
+def zeros_like(a, dtype=None, shape=None):
+    result = torch.zeros_like(a, dtype=dtype)
+    if shape is not None:
+        result = result.reshape(shape)
```
lezcano, line 436: Fantastic candidate to be implemented via a decorator

lezcano (approved):

torch_np/_detail/implementations.py
ev-br: Let's keep this in mind for future refactors.

ev-br:
Thanks Mario for the review.

PR #53, opened 2023-02-12
Push dtypes to torch, move scalar types 

ev-br:
Make the split between torch.dtypes and wrapper ndarray dtypes cleaner, move scalar_types to the the public area, merge with wrapper dtypes --- the idea of having scalar types being both public and used for dtype decisions internally does not play out in the end, so ditch it.
Overall the current version feels right while what is in main does not, FWIW.


lezcano (approved):
Great! Now the imports seem to be going just in one direction, as one would expect! :)

ev-br:
Yup, chiseling it bit by bit :-).

PR #54, opened 2023-02-12
CI: use pytest-xdist with 2 cores on GHA

ev-br:
IIUC, GH hosted runners have two cores (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners), so use them.


ev-br:
CI is green, and this seems to shave off about 50-60 seconds from the test run. Merging.

ev-br:
Thanks @honno, merged. I've no prior experience with pytest-xdist either, other that here it seems to work locally and on CI. Let's keep an eye on if it causes any problems

PR #55, opened 2023-02-15
Run test CI on pushes for any branch and PR activity

honno:
Currently our test CI only runs for `main` changes, so it'd be nice to get it running for any other branches, and when folks open/push-to PRs generally.


ev-br:
> folks open/push-to PRs generally.

 Hear hear :-)

PR #57, opened 2023-02-16
BUG: fix assert_equal for nones and dtypes

ev-br:


ev-br:
A minimal fix for a bug in testing which was masking actual bugs in implementations. Of the latter, fix a low hanging fruit,  xfail things which need a rework of the ufunc casting machinery---which I'm kicking down the road for after there is an MVP.

lezcano (approved):

PR #62, opened 2023-02-22
add ndarrays.flags.f_contiguous

ev-br:
Implement `f_contguous` flag, also heuristics for `owndata` and `writeable` flags.

### DIFF
--- a/torch_np/_ndarray.py
+++ b/torch_np/_ndarray.py
@@ -115,12 +115,18 @@ class ndarray:
 
         # check if F contiguous
         from itertools import accumulate
-        f_strides = tuple(accumulate(list(self._tensor.shape), func=lambda x, y: x*y))
+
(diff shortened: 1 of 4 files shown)


torch_np/_ndarray.py
```diff
@@ -112,7 +112,22 @@ def base(self):
     @property
     def flags(self):
         # Note contiguous in torch is assumed C-style
-        return Flags({"C_CONTIGUOUS": self._tensor.is_contiguous()})
+
+        # check if F contiguous
+        from itertools import accumulate
+
+        f_strides = tuple(accumulate(list(self._tensor.shape), func=lambda x, y: x * y))
```
ev-br, line 119: this is of course just `torch.cumprod(self._tensor.shape, 0)`; my instinct from numpy usage is to avoid array operations for small tuples etc.

lezcano:
if things are implemented in the wrapper, we can certainly use those to implement others as those will redirect ultimately ti PyTorch ops, so that's completely fine

PR #63, opened 2023-02-22
MAINT: work around arg{min,max} not implemented for booleans

ev-br:


torch_np/_detail/_reductions.py
```diff
@@ -36,12 +36,22 @@ def count_nonzero(a, axis=None):
 
 def argmax(tensor, axis=None):
     axis = _util.allow_only_single_axis(axis)
+
+    if tensor.dtype == torch.bool:
+        # RuntimeError: "argmax_cpu" not implemented for 'Bool'
+        tensor = tensor.view(torch.int8)
```
ev-br, line 42: Is the equivalence guaranteed in pytorch?

lezcano: I would cast it to `uint8` to be 100% sure, but yeah, I believe it is guaranteed by some work @peterbell10 did some time ago.

ev-br:
By cast you mean tensor.to or tensor.view?

lezcano:
view is fine.

torch_np/_detail/_reductions.py
peterbell10: A cast is guaranteed to be 0 or 1, but a view could contain any byte value.

peterbell10: Oh and of course the work I did doesn't apply to the `torch.compile` stack.

ev-br:
Based on https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/63#discussion_r1114856264 and https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/63#discussion_r1114739469 let's copy to uint8 to be on the safe side.

torch_np/_detail/_reductions.py
lezcano: In this case, if viewing with `uint8` we would get something that's non zero, so argmax would then work, but sure, let's do `to`, as the compiler should be able to optimise out the copy really

peterbell10: IIRC argmax always returns the lowest index that's equal to the maximum, so the distinction does matter.

PR #70, opened 2023-02-27
bare-bones normalizations via type hints

ev-br:
An experiment along the lines of the suggestion in https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/32 : (ab)use type annotations to mark the logic for normalizing user-facing arguments into pytorch tensors and dtypes, reshuffle arguments etc via static types.

A somewhat clumsy bit is normalizing `*args` --- case in point is e.g. `atleast_1d(*arys)`.
The issue is that `func(*args : Annotation)` gives a single annotation for a runtime-determined number of arguments. There is no way to annotate individual elements of *args AFAICS.

Thus register a special annotation to repack args into a tuple and a corresponding normalizer to normalize this tuple.


ev-br:
The initial scaffolding it these, so am removing the WIP admonition. While this is not yet complete (ufuncs, reductions), this is something to take a look at, if only to see if a general pattern is tolerable.

torch_np/_detail/_reductions.py
```diff
@@ -11,6 +11,46 @@
...
+
+
+def deco_axis_expand(func):
+    """Generically handle axis arguments in reductions."""
+    @functools.wraps(func)
+    def wrapped(tensor, axis, *args, **kwds):
+
+        if axis is not None:
+            if not isinstance(axis, (list, tuple)):
+                axis = (axis,)
```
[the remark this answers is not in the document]
ev-br: Thanks @honno

ev-br:
This is ready from my side.

Further possible cleanups include 

- [ ] annotate the return type. This could centralize the majority of `_helpers.array_from / tuple_arrays_from` incantations, and `result_or_out`.
- [ ] possibly centralize normalization/validation of axis arguments.

both are somewhat large and need some experimentation, so are best postponed to follow-up PRs.

lezcano:
Just passing by to say that the output one can be done without anotation. Just do a `if isintance(out, torch.Tensor): return as_array(out)` in the normalization wrapper, and that'll get rid of all those `_helper.array_from`. You can do similarly for tuples. This can be done because a function's returned value is already "normalised" (it'll always be a tensor or a tuple of tensors or things that we don't want to touch).

For the `out=` kwarg, we can do it like in PrimTorch:
https://github.com/pytorch/pytorch/blob/ab148da66cb9433effac90c7bd4930a961481d19/torch/_prims_common/wrappers.py#L187
Note that that code is quite tricky, because it handles namedtuples and it also puts in the correct annotation, but you can probably simplify it in our case.

ev-br:
> Just do a if isintance(out, torch.Tensor): return as_array(out)

Hmm, this breaks in presence of `out=` argument unless `as_array` has the `out=` semantics?

lezcano:
But if we implement the `out=` kwarg ourselves as a decorator, we should be able make both things work with each other.

ev-br:
Certainly can. My point is simply that it's a bit more than just using `if isinstance(result, torch.Tensor): return asarray(result)`, and I'd prefer to limit the scope of this PR and deal with returns in a follow-up.

lezcano:
About annotating variadic args, does this SO solve the question? https://stackoverflow.com/a/37032111/5280578

torch_np/_helpers.py
```diff
@@ -70,7 +81,7 @@ def result_or_out(result_tensor, out_array=None, promote_scalar=False):
         out_tensor.copy_(result_tensor)
         return out_array
     else:
-        return asarray(result_tensor)
+        return array_from(result_tensor)
```
lezcano, line 84: Why this change?

torch_np/_normalizations.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def normalize_ndarray(arg, name=None):
+    if arg is None:
+        return arg
+
+    from ._ndarray import ndarray
+
+    if not isinstance(arg, ndarray):
+        raise TypeError("'out' must be an array")
```
lezcano, line 71: If this is just used for the `out=` parameter, let's call it accordinly.

torch_np/_normalizations.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def normalize_this(arg, parm):
+    """Normalize arg if a normalizer is registred."""
+    normalizer = normalizers.get(parm.annotation, None)
+    if normalizer:
+        return normalizer(arg)
+    else:
+        # untyped arguments pass through
+        return arg
```
lezcano, line 96: If you prefer, this can be a one-liner:
```suggestion
    return normalizers.get(param.annotation, lambda x: x)(arg)
```

torch_np/tests/test_ndarray_methods.py
```diff
@@ -17,7 +17,7 @@ def test_indexing_simple(self):
 
         assert isinstance(a[0, 0], np.ndarray)
         assert isinstance(a[0, :], np.ndarray)
-        assert a[0, :].base is a
```
lezcano, line 20: Are these changes coming from a weird ordering of the stacks or what'st hte reason?

torch_np/_detail/_reductions.py
```diff
@@ -4,13 +4,78 @@
...
+                if not isinstance(axis, typing.SupportsIndex):
+                    raise TypeError(
+                        f"{type(axis)=}, but should be a list/tuple or support operator.index()"
+                    )
+                axis = (axis,)
+            axis = _util.normalize_axis_tuple(axis, tensor.ndim)
+
+        if axis == ():
+            newshape = _util.expand_shape(tensor.shape, axis=0)
+            tensor = tensor.reshape(newshape)
```
lezcano, line 39: I had to look up what `expand_shape` does here. But doesn't this simply do
```suggestion
            tensor = tensor.reshape((1,))
```

torch_np/_detail/_reductions.py
```diff
@@ -4,13 +4,78 @@
...
+            axis = (0,)
+
+        result = func(tensor, axis=axis, *args, **kwds)
+        return result
+
+    return wrapped
+
+
+def emulate_keepdims(func):
```
lezcano, line 48: This should probably be on the main normalization suit. Then, I reckon it should be alright to simply inline `apply_keepdims` there.

torch_np/_detail/_unary_ufuncs.py
```diff
@@ -0,0 +1,55 @@
+"""Export torch work functions for unary ufuncs, rename/tweak to match numpy.
+This listing is further exported to public symbols in the `torch_np/_unary_ufuncs.py` module.
+"""
+
+import torch
+
+# renames
```
lezcano, line 7: Why the different imports here, rather than just do a `from torch import ( blablalba )`?

torch_np/_wrapper.py
```diff
@@ -445,48 +495,52 @@ def size(a, axis=None):
...
 
 # YYY: pattern: tuple of arrays as input, tuple of arrays as output; cf nonzero
-def broadcast_arrays(*args, subok=False):
-    _util.subok_not_ok(subok=subok)
-    tensors = _helpers.to_tensors(*args)
-    res = torch.broadcast_tensors(*tensors)
-    return tuple(asarray(_) for _ in res)
+@normalizer
+def broadcast_arrays(*args: UnpackedSeqArrayLike, subok: SubokLike = False):
+    args = args[0]  # undo the *args wrapping in normalizer
```
lezcano, line 541: We should hopefully be able to fix this with a proper preprocessing of variadic inputs in the normalizer.

lezcano (commented):
`torch_np/_wrapper` starts looking good! Let's revisit the idea of moving the actual implementations of the torch functions to the wrapper after we have merged this PR and the return PR.

torch_np/_wrapper.py
```diff
@@ -4,14 +4,22 @@
...
 
-from . import _funcs
+from . import _decorators, _dtypes, _funcs, _helpers
 from ._detail import _dtypes_impl, _flips, _reductions, _util
 from ._detail import implementations as _impl
-from ._ndarray import array, asarray, asarray_replacer, maybe_set_base, ndarray, newaxis
-
-from . import _dtypes, _helpers, _decorators  # isort: skip  # XXX
+from ._ndarray import array, asarray, maybe_set_base, ndarray
+from ._normalizations import (
```
lezcano, line 15: What's the difference between `_funcs.py` and `_wrapper.py`? Should we merge them?

torch_np/_detail/_unary_ufuncs.py
ev-br: This is `isort` from `pre-commit run`.

torch_np/_wrapper.py
ev-br: Absolutely. The split is temporary, and we'll cleanly merge them once the cleanups from other PRs further up the stack are done. 
At this stage, imports from ._ndarray cause circular imports. So either let's live with the split for a while, or I can bloat this PR with what's up the stack. Reviewer's choice :-).

torch_np/tests/test_ndarray_methods.py
ev-br: Can remove it from this one if wanted.

torch_np/_helpers.py
ev-br: Purely technical. Not a big deal in this file (just makes it easier to grep / more uniform). 
The general direction is centralizing imports from `_ndarray.py` to minimize/centralize local imports.

torch_np/_helpers.py
```diff
@@ -29,18 +29,33 @@ def cast_and_broadcast(tensors, out, casting):
     if out is None:
         return tensors
     else:
-        from ._ndarray import asarray, ndarray
-
-        if not isinstance(out, ndarray):
-            raise TypeError("Return arrays must be of ArrayType")
-
         tensors = _util.cast_and_broadcast(
             tensors, out.dtype.type.torch_dtype, out.shape, casting
```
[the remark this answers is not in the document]
ev-br: Note that this whole function is inlined/removed in https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/89 (note to self to check the test coverage)

torch_np/_detail/_reductions.py
ev-br: `keepdims` interacts with `axis`, so moving this to the main normalization dance would break an invariant that normalization of an argument only depends on this argument and nothing else.

torch_np/_helpers.py
```diff
@@ -29,18 +29,33 @@ def cast_and_broadcast(tensors, out, casting):
...
+    if dtype is not None:
+        raise NotImplementedError
+
+    out_shape_dtype = None
+    if out is not None:
+        out_shape_dtype = (out.get().dtype, out.get().shape)
+
+    tensors = _util.cast_and_broadcast(tensors, out_shape_dtype, casting)
+
+    return tensors
```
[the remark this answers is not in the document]
ev-br: Not quite. In numpy, `out` array broadcasts ufunc inputs, and is never resized:

```
In [1]: import torch

In [2]: t = torch.ones(2, dtype=int)

In [3]: out = torch.empty((2, 2))

In [4]: res = torch.add(t, t, out=out)
<ipython-input-4-1158967e8a80>:1: UserWarning: An output with one or more elements was resized since it had shape [2, 2], which does not match the required output shape [2]. This behavior is deprecated, and in a future PyTorch release outputs will not be resized unless they have zero elements. You can explicitly reuse an out tensor t by resizing it, inplace, to zero elements with t.resize_(0). (Triggered internally at /home/conda/feedstock_root/build_artifacts/pytorch-recipe_1670076225403/work/aten/src/ATen/native/Resize.cpp:17.)
  res = torch.add(t, t, out=out)

In [5]: res
Out[5]: tensor([2., 2.])

In [6]: import numpy as np

In [7]: np.add(t.numpy(), t.numpy(), out=np.empty((2, 2)))
Out[7]: 
array([[2., 2.],
       [2., 2.]])
```

torch_np/_detail/_reductions.py
ev-br: Not really. 

```
In [14]: t = torch.ones(2, dtype=int)

In [15]: t.reshape((1,))
Input In [15], in <cell line: 1>()
----> 1 t.reshape((1,))

RuntimeError: shape '[1]' is invalid for input of size 2

In [16]: from torch_np._detail import _util

In [17]: _util.expand_shape(t.shape, axis=0)
Out[17]: [1, 2]
```

What's going on here is that `expand_shape` is numpy's `expand_dims`, basically `torch.unsqueeze`. Here we insert a length-one axis and run the reduction along it. This is to emulate this numpy's beauty:

```
In [19]: np.sum(t.numpy(), axis=())
Out[19]: array([1, 1])
```

which is different from 

```
In [24]: np.sum(t.numpy(), axis=None)
Out[24]: 2
```

torch_np/_normalizations.py
```diff
@@ -0,0 +1,164 @@
...
+        ba = sig.bind(*lst, **dct)
+        ba.apply_defaults()
+
+        # Now that all parameters have been consumed, check:
+        # Anything that has not been bound is unexpected positional arg => raise.
+        # If there are too few actual arguments, this fill fail in func(*ba.args) below
+        if len(args) > len(ba.args):
+            raise TypeError(
+                f"{func.__name__}() takes {len(ba.args)} positional argument but {len(args)} were given."
+            )
```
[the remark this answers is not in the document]
ev-br: OK, done in https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/70/commits/8c78725def3682bfca3036a23c955241cfac7f67.

torch_np/_helpers.py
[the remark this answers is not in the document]
lezcano: ugh, that's incredibly annoying.

torch_np/tests/test_ndarray_methods.py
lezcano: My point is, these are unrelated to this PR as, I believe, the `base` arg was removed in a different PR. As such, what are they doing here?

torch_np/_helpers.py
lezcano: makes sense, and thank you for the context!

torch_np/_helpers.py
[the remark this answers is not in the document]
lezcano: Actually, can't we simply do as we do in PyTorch but first broadcasting the output of the function to the size of the `out=` tensor before performing the `copy_`? That would allow us to implement the `out=` semantics separately from all the rest, which should be much simpler. This uses the (trivial) fact that broadcasting commutes with pointwise ops.

torch_np/_detail/_reductions.py
lezcano: Right. Thank you for the comment!

torch_np/_detail/_unary_ufuncs.py
lezcano: Terribly weird. @honno any idea what's going on here?

torch_np/_normalizations.py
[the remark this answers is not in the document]
lezcano: Great! gotta love functional programming.

torch_np/_helpers.py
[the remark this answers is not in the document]
ev-br: We'll still have to manually do type casting in the preprocessing though. Specifically, this loop, https://github.com/Quansight-Labs/numpy_pytorch_interop/blob/main/torch_np/_detail/_util.py#L198 cannot be avoided, no matter where precisely in lands.

Whether this is going to be much simpler than https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/89 or not, I frankly don't know.  Where is the expected simplification I'm not seeing?

torch_np/_detail/_reductions.py
lezcano: I get the feeling that it should be possible to simplify these further. Let's leave as it is for now though. I may put some more thought on this at a later time.

torch_np/_helpers.py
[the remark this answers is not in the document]
ev-br: How does it show it, mind elaborating?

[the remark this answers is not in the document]
ev-br: Ah, thanks. I think it's the same thing as these tests in gh-89: https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/89/files

I believe (cf comments in tests) what we have now is a superset of what numpy is doing, esp for binary ufuncs.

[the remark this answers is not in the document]
ev-br: Also, here's one more snippet of why we need at least some pre-processing:

```
In [31]: np.sin(True, out=np.empty(1, dtype=bool))
Input In [31], in <cell line: 1>()
----> 1 np.sin(True, out=np.empty(1, dtype=bool))

UFuncTypeError: Cannot cast ufunc 'sin' output from dtype('float32') to dtype('bool') with casting rule 'same_kind'

In [32]: np.sin(True, out=np.empty(1, dtype=bool), casting='unsafe')
Out[32]: array([ True])
```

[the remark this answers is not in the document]
lezcano: I don't think that that example proves we need preprocessing. The semantics for those two examples are:
- Run he function
- Check whether the output dtype can be cast to the dtype of `out`
- If not, check whether `casting="unsafe"` otherwise raise an error
And all this is done after running the function.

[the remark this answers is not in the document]
ev-br: First, there are more casting modes:

```
In [39]: np.sin(True, out=np.empty(1, dtype=bool), casting='no')
Input In [39], in <cell line: 1>()
----> 1 np.sin(True, out=np.empty(1, dtype=bool), casting='no')

UFuncTypeError: Cannot cast ufunc 'sin' input from dtype('bool') to dtype('float32') with casting rule 'no'
```

Second, what we are discussing now is I believe a (bunch of) corner case(s). What's more interesting is binary ufuncs. For one,

```
In [41]: np.add(True, 2, casting='no')
Input In [41], in <cell line: 1>()
----> 1 np.add(True, 2, casting='no')

UFuncTypeError: Cannot cast ufunc 'add' input 0 from dtype('bool') to dtype('int64') with casting rule 'no'
```

torch_np/_normalizations.py
```diff
@@ -0,0 +1,164 @@
...
+        for j, param in enumerate(sig.parameters.values()):
+            if param.annotation == UnpackedSeqArrayLike:
+                if j == 0:
+                    args = (args,)
+                else:
+                    # args = args[:j] + (args[j:],) would likely work
+                    # not present in numpy codebase, so do not bother just yet.
+                    # NB: branching on j ==0 is to avoid the empty tuple, args[:j]
+                    raise NotImplementedError
+                break
```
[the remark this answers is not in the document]
ev-br: That won't work, because `tnp.atleast(1, 2)` gets two arguments and only a single annotation. I can of course check `param.kind == VAR_POSITIONAL` instead of a special annotation and consume the rest of `args` instead of checking `if param.annotation == UnpackedSeqArrayLike`, this does not seem any simpler or clearer, does it.

torch_np/_wrapper.py
ev-br: Done in https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/70/commits/9d75cab63adaad254996e27f4a927ee4591d0ff0

torch_np/_normalizations.py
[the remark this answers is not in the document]
ev-br: OK, how about https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/70/commits/9d75cab63adaad254996e27f4a927ee4591d0ff0

torch_np/_helpers.py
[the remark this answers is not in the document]
lezcano: All those other casting modes should also be performed on the result, not on the out args. The casting args may affect teh casting of the inputs, but they affect them irregardless of whether the `out=` kwarg is present or not.
Let's have a vc to discuss this on Wed, I think it'll be easier.

torch_np/_wrapper.py
lezcano: Ah, right, I forgot we need to error out on those. In my head I always thought of the non-error case.

Also, I appreciate it, but no need to attribute the commit really :)

torch_np/_normalizations.py
```diff
@@ -0,0 +1,159 @@
...
+        # NumPy's API does not have positional args before variadic positional args
+        if first_param.kind == inspect.Parameter.VAR_POSITIONAL:
+            args = [maybe_normalize(arg, first_param) for arg in args]
+        else:
+            args = [
+                maybe_normalize(arg, parm) for arg, parm in zip(args, params.values())
+            ]
+
+        # NB: extra unknown arguments: pass through, will raise in func(*args) below
+        args += args_copy[len(args) :]
```
lezcano, line 106: Just a teeny tiny thing to avoid having to copy the args, which is not very efficient
```suggestion
        # NumPy's API does not have positional args before variadic positional args
        if first_param.kind == inspect.Parameter.VAR_POSITIONAL:
            args = [maybe_normalize(arg, first_param) for arg in args]
        else:
            # NB: extra unknown arguments: pass through, will raise in func(*args) below
            args = [
                maybe_normalize(arg, parm) for arg, parm in zip(args, params.values())
            ] + args[len(params.values():]
```

lezcano (approved):
Let's avoid copying the args, but otherwise this LGTM. As discussed, let's merge all the PRs as-is (as-are?) and let's try to factor out the `out=` kwarg implementation and then have a simple implementation of the wrapping of the outputs.

PR #72, opened 2023-02-28
Add ndarray.dot

ev-br:
Now that both np.dot and ndarray.dot are there, can remove xfails in tests. Which, in turn, smokes out several problems in arange(..., dtype=complex), so fix them. And while at it, remove xfails of TestArange itself and fix it, too :-)

### DIFF
--- a/torch_np/_detail/_ufunc_impl.py
+++ b/torch_np/_detail/_ufunc_impl.py
@@ -1,6 +1,7 @@
 import torch
 
 from . import _util
+from . import _dtypes_impl
 
 
 def deco_ufunc(torch_func):
@@ -70,7 +71,6 @@ logaddexp2 = deco_ufunc(torch.logaddexp2)
 logical_and = deco_ufunc(torch.logical_and)
 logical_or = deco_ufunc(torch.logical_or)
 logical_xor = deco_ufunc(torch.logical_xor)
-matmul = deco_ufunc(torch.matmul)
 maximum = deco_ufunc(torch.maximum)
 minimum = deco_ufunc(torch.minimum)
 remainder = deco_ufunc(torch.remainder)
@@ -143,7 +143,15 @@ def _absolute(x):
         return x
     return torch.absolute(x)
 
+def _matmul(x, y):
+    # work around RuntimeError: expected scalar type Int but found Double
+    dtype = _dtypes_impl.result_type_impl((x.dtype, y.dtype))
+    x = x.to(dtype)
+    y = y.to(dtype)
+    result = torch.matmul(x, y)
+    return result
 
 cbrt = deco_ufunc(_cbrt)
 positive = deco_ufunc(_positive)
 absolute = deco_ufunc(_absolute)
+matmul = deco_ufunc(_matmul)
--- a/torch_np/_detail/_ufunc_impl.py
+++ b/torch_np/_detail/_ufunc_impl.py
@@ -1,7 +1,6 @@
 import torch
 
-from . import _util
-from . import _dtypes_impl
+from . import _dtypes_impl, _util
 
 
 def deco_ufunc(torch_func):
@@ -143,6 +142,7 @@ def _absolute(x):
         return x
     return torch.absolute(x)
 
+
 def _matmul(x, y):
     # work around RuntimeError: expected scalar type Int but found Double
     dtype = _dtypes_impl.result_type_impl((x.dtype, y.dtype))
@@ -150,6 +150,7 @@ def _matmul(x, y):
     y = y.to(dtype)
     result = torch.matmul(x, y)
     return result
+
 
 cbrt = deco_ufunc(_cbrt)
 positive = deco_ufunc(_positive)
--- a/torch_np/_detail/_ufunc_impl.py
+++ b/torch_np/_detail/_ufunc_impl.py
@@ -146,8 +146,8 @@ def _absolute(x):
 def _matmul(x, y):
     # work around RuntimeError: expected scalar type Int but found Double
     dtype = _dtypes_impl.result_type_impl((x.dtype, y.dtype))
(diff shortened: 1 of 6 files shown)


torch_np/_ndarray.py
```diff
@@ -362,6 +362,7 @@ def reshape(self, *shape, order="C"):
 
     diagonal = _funcs.diagonal
     trace = _funcs.trace
+    dot = _funcs.dot
```
lezcano, line 365: missing also `vdot`?

torch_np/_funcs.py
```diff
@@ -93,7 +93,20 @@ def fill_diagonal(a, val, wrap=False):
...
+    t_a, t_b = _helpers.to_tensors(a, b)
+    result = _impl.vdot(t_a, t_b)
+    return result.item()
+
+
+def dot(a, b, out=None):
+    t_a, t_b = _helpers.to_tensors(a, b)
+    dtype = _dtypes_impl.result_type_impl((t_a.dtype, t_b.dtype))
+    t_a = t_a.to(dtype)
+    t_b = t_b.to(dtype)
```
lezcano, line 106: cast_if_needed

lezcano: Also, why isn't the type cast within `_impl.dot`?

torch_np/_detail/implementations.py
```diff
@@ -492,17 +492,31 @@ def arange(start=None, stop=None, step=1, dtype=None):
...
+
+    # work around RuntimeError: "arange_cpu" not implemented for 'ComplexFloat'
+    if dtype.is_complex:
+        work_dtype, target_dtype = torch.float64, dtype
+    else:
+        work_dtype, target_dtype = dtype, dtype
+
+    if (step > 0 and start > stop) or (step < 0 and start < stop):
+        # empty range
+        return torch.as_tensor([], dtype=target_dtype)
```
lezcano, line 510:
```suggestion
        return torch.empty(0, dtype=target_type)
```

lezcano (approved):
A few minor points, but overall LGTM

torch_np/_ndarray.py
ev-br: Missing in numpy, yes :-)

In [23]: hasattr(np.array([1, 2, 3]), 'vdot')
Out[23]: False

torch_np/_funcs.py
ev-br: Thanks, indeed. Done,  x2

torch_np/_detail/_ufunc_impl.py
```diff
@@ -144,6 +143,16 @@ def _absolute(x):
     return torch.absolute(x)
 
 
+def _matmul(x, y):
+    # work around RuntimeError: expected scalar type Int but found Double
+    dtype = _dtypes_impl.result_type_impl((x.dtype, y.dtype))
+    x = x.to(dtype)
+    y = y.to(dtype)
+    result = torch.matmul(x, y)
+    return result
```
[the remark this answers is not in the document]
ev-br: Devil's in the details (edge cases). I'll see if implementations can be merged, after all wrinkles are ironed out. For now there are xfails still.

From https://numpy.org/doc/stable/reference/generated/numpy.matmul.html

```
matmul differs from dot in two important ways:

Multiplication by scalars is not allowed, use * instead.

Stacks of matrices are broadcast together as if the matrices were elements, respecting the signature (n,k),(k,m)->(n,m):
```

torch_np/_ndarray.py
lezcano: lol. Let's still add it to be forward-looking.

torch_np/_detail/_ufunc_impl.py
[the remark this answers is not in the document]
lezcano: wow, ok, then `torch.matmul` is `np.dot`. You'll need to implement `np.matmul` implementing that weird broadcasting behaviour by hand.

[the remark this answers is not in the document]
ev-br: Yes (sigh). Am going to postpone this  a bit in favor of gh-70.  Once that stabilizes, will turn back to matmul.

torch_np/_ndarray.py
ev-br: OK, can do. 
If we're starting to do this, what's the guiding principle, where do we stop however. E.g. do we want feature parity for the main namespace and ndarray methods?

```
In [24]: len(dir(np))
Out[24]: 595

In [25]: len(dir(np.ndarray))
Out[25]: 165
```

torch_np/_detail/_ufunc_impl.py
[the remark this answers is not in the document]
lezcano: I don't quite understand what you mean by "this signature thing", but I happen to know that PyTorch doesn't have generalised ufuncs, so we'll need to replicate that machinery at some point.

torch_np/_ndarray.py
lezcano: If we can do it "for free" it may be a fine thing to do? WDYT @rgommers?

torch_np/_detail/_ufunc_impl.py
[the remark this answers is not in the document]
ev-br: "This signature thing":

> respecting the signature (n,k),(k,m)->(n,m)

also referred to as _that weird broadcasting_ a couple of messages above :-).

So yes, am going to postpone this to some point in the future.

torch_np/_ndarray.py
rgommers: There is no guiding principle, it's an accident of history. It's highly unlikely though that we'll add more method to `numpy.ndarray`, there are already too many. So I wouldn't add anything that's not already present.

PR #75, opened 2023-03-01
Allow package installation

larryliu0820:
Adding `pyproject.toml` and `setup.py`. Version is hardcoded and should probably change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,13 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "numpy_pytorch_interop"
+dependencies = [
+    "numpy",
+    "torch",
+]
+version = "0.1.0"
+[tool.setuptools.packages.find]
+include = ["torch_np.*"]  # ["*"] by default
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,3 @@
+from setuptools import setup
+
+setup()


larryliu0820:
@ev-br @honno @lezcano mind taking a look?

pyproject.toml
```diff
@@ -0,0 +1,15 @@
...
+[project]
+name = "numpy_pytorch_interop"
+dependencies = [
+    "numpy",
+    "torch",
+]
+version = "0.1.0"
+[tool.setuptools.packages.find]
+where = ["."]
+include = ["autogen", "torch_np.*"]  # ["*"] by default
```
ev-br, line 14: autogen folder is not really needed, there's nothing worth installing really

pyproject.toml
```diff
@@ -0,0 +1,15 @@
...
+exclude = ["torch_np.tests*", "torch_np.testing"]  # empty by default
```
ev-br, line 15: Why?

ev-br (commented):
As a pure python package,  just setting PYTHONPATH is technically enough. Could you elaborate on how this makes what easier or more convenient? 

If it's needed for something, I'm on a fence tbh, so can merge and roll on, my comments can be easily addressed in a follow-up

larryliu0820:
> As a pure python package, just setting PYTHONPATH is technically enough. Could you elaborate on how this makes what easier or more convenient?
> 
> If it's needed for something, I'm on a fence tbh, so can merge and roll on, my comments can be easily addressed in a follow-up

Yeah I thought it might be easier to install than manually setting PYTHONPATH. I don't mind setting PYTHONPATH though. Also is there a plan to publish/distribute this project to, say pypi?

pyproject.toml
larryliu0820: updated.

ev-br:
> Also is there a plan to publish/distribute this project to, say pypi?

Good question. Do you think it would be useful as a stand alone package? I'm genuinely curious on your intended use

larryliu0820:
> > Also is there a plan to publish/distribute this project to, say pypi?
> 
> Good question. Do you think it would be useful as a stand alone package? I'm genuinely curious on your intended use

I'm working on a pytorch issue where we want to use `torch_np` in pytorch compiler. See https://github.com/pytorch/pytorch/issues/93684#issuecomment-1436026703 for more details. In the PR I'm working on, I want to check if `numpy_pytorch_interop` is installed:

"""python
try:
    import torch_np
    HAS_NUMPY_TORCH_INTEROP = True
except ModuleNotFoundError:
    torch_np = None
    HAS_NUMPY_TORCH_INTEROP = False
"""

ev-br:
Thanks @larryliu0820 , merged.

ev-br:
Great to hear you're working on that! Feel free to ping me if there are glitches (there surely are).

PR #76, opened 2023-03-02
Introduce `--nonp` to error when NumPy is accessed

honno:
Using `--nonp` will error before any tests are run when NumPy is accessed. I think this is especially useful as we can use it iteratively—
```console
$ cat torch_np/tests/test_foo.py
def test_foo:
    assert True
$ pytest torch_np/tests/test_foo.py
# no error
```

No error is raised because the tests pytest has collected never import NumPy. Meanwhile 

```console
$ cat torch_np/_uses_numpy.py
import numpy as np
...
np.int8
...
$ cat torch_np/tests/test_bar.py
from torch_np import _uses_numpy
$ pytest torch_np/tests/test_bar.py
...
ERROR torch_np/tests/test_bar.py - RuntimeError: Using --nonp but accessed np.int8
    torch_np/tests/test_bar.py:1: in <module>
        ???
    torch_np/_uses_numpy.py:4: in <module>
        ???
    torch_np/tests/conftest.py:20: in __getattribute__
        ???
    E   RuntimeError: Using --nonp but accessed np.int8
...
ERROR torch_np/tests/test_bar.py - RuntimeError: Using --nonp but accessed np.int8
```

i.e. only when the test ends up having NumPy imported does it error on the collection stage. This means we can leverage the coverage of individual test files and work on getting those passing before jumping on to another, til eventually just `pytest torch_np/tests/ --nonp` passes (and then we can introduce a CI job for it to prevent regressions).

### DIFF
--- a/torch_np/tests/conftest.py
+++ b/torch_np/tests/conftest.py
@@ -1,3 +1,5 @@
+import sys
+
 import pytest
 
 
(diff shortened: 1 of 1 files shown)


ev-br:
This looks slick, wow :-)

PR #77, opened 2023-03-02
Support testing `np` and `tnp` on the same test

honno:
@ev-br here's how one could test both NumPy-proper and `torch_np` on the same test. Essentially any test case just needs a parameter `np`, which should be treated as the array module we want to test (i.e. `numpy`, `torch_np`), and we need to ensure we're using that module rather than outside imports.

I reworked some of `test_indexing.py` to demonstrate how this could be utilised. It's a bit finnicky as you have to check `np is {numpy,torch_np}` for some module-specific behaviour, e.g. raised error classes and skips/xfails, and you have to make sure you're using the argument `np` for things like the `{numpy,torch_np}.testing` utils.

Maybe we could support `np` as a magically param but not really bother to rework existing tests with it—instead we use it as-and-when we feel it useful to "sanity check" a test works the same for NumPy. So IMO it might be worth merging just the `conftest.py` changes.

### DIFF
--- a/torch_np/tests/conftest.py
+++ b/torch_np/tests/conftest.py
@@ -2,6 +2,8 @@ import sys
 
 import pytest
 
+import torch_np as tnp
+
 
 def pytest_configure(config):
     config.addinivalue_line("markers", "slow: very slow tests")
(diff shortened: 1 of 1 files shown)


ev-br:
Merging conftest.py change (+explanation, this looks rather black magic to me at least :-)) LGTM. The rest, let's wait for Mario to weight in?

ev-br:
Thank you @honno! 
Let's track desired patterns of actually using this feature in https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/78

PR #79, opened 2023-03-08
Fix pyproject.toml

larryliu0820:
Include both files and subdirectory under `torch_np`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,4 +10,4 @@ dependencies = [
 ]
 version = "0.1.0"
 [tool.setuptools.packages.find]
-include = ["torch_np.*"]  # ["*"] by default
+include = ["torch_np", "torch_np.*"]  # ["*"] by default


honno:
Weird, yeah the previous pyproject won't build the files directly under `torch_np/` for the wheel, but will for the tarball (when using `python -m build`). LGTM

@larryliu0820 out of curiousity, why would one include a `setup.py` if using `pyproject.toml`?

larryliu0820:
> Weird, yeah the previous pyproject won't build the files directly under `torch_np/` for the wheel, but will for the tarball (when using `python -m build`). LGTM
> 
> @larryliu0820 out of curiousity, why would one include a `setup.py` if using `pyproject.toml`?

Yeah `pyproject.toml` is still in beta,  for some old environment it may not recognize so need a dummy `setup.py`.

larryliu0820:
@honno can you please help merge this? I don't have access.

honno:
Trivial change that's nice to get in, so merging.

> Yeah `pyproject.toml` is still in beta, for some old environment it may not recognize so need a dummy `setup.py`.

Makes sense!

Issue #47, opened 2023-02-07
ndarray.base : worth keeping?


ev-br:
NumPy uses the `.base` ndarray attribute to trace the views, as in 

```
In [77]: a = np.arange(10)

In [78]: a[::2].base is a
Out[78]: True

In [79]: a[::2].copy().base is a
Out[79]: False
```

It's been suggested (https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/46#discussion_r1098952569) to drop this, because PyTorch tracks views vs copies in a private `Tensor._base` attribute.

While there is no way to restore the pubic `base` from private `_tensor._base`, the question is whether we want/need to keep the `.base` on the ndarray wrapper level.

PR #80, opened 2023-03-11
API: remove ndarray.base

ev-br:
closes gh-47

Base tensor relationship is tracked via pytorch'es `tensor._base` already, so remove  `ndarray.base` attribute.
Checking an wrapper array base becomes then

```
>>> a.get()._base is b.get()
```

instead of numpy's

```
>>> a.base is b
```

This is the second PR in the "stack" based off gh-70.

EDIT : a.get() is gone, so it's `a.tensor._base`  now. 

closes gh-47

### DIFF
--- a/torch_np/_decorators.py
+++ b/torch_np/_decorators.py
@@ -17,7 +17,7 @@ def out_shape_dtype(func):
     @functools.wraps(func)
     def wrapped(*args, out=None, **kwds):
         if out is not None:
-            kwds.update({"out_shape_dtype": (out.get().dtype, out.get().shape)})
+            kwds.update({"out_shape_dtype": (out.tensor.dtype, out.tensor.shape)})
         result_tensor = func(*args, **kwds)
         return _helpers.result_or_out(result_tensor, out)
 
--- a/torch_np/_dtypes.py
+++ b/torch_np/_dtypes.py
@@ -49,7 +49,7 @@ class generic(abc.ABC):
         # and here we follow the second approach and create a new object
         # *for all inputs*.
         #
-        return _ndarray.ndarray._from_tensor_and_base(tensor, None)
+        return _ndarray.ndarray._from_tensor(tensor)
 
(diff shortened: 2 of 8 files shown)


lezcano (commented):
In this clean-up, we can also remove the `.get()` method (now it's just `._tensor` (or `.tensor` if you want, no need to make that member private, it's a bit unpythonic IMO) and `ndarray._from_tensor(x)` can be merged into the `__init__` so that we can write the more natural `ndarray(t)`.

ev-br:
FWIW, I'd rather keep only having an argument-less `__init__`  or maybe even no `__init__` at all, to avoid mixing up with https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html which is very much not recommended for a long time now but is still used in old scripts here and there. 

W.r.t. public or private `tensor` attribute, I suppose my instincts are from a library maintainer perspective : anything which is not explicitly private will get (ab)used in the wild and will cause backcompat issues down the line.

lezcano:
Note that this ndarray is not a public API but an intermediary artefact used by a compiler, so we should write it in whichever way we find easier. 
Same for `__init__` really. Note that it makes sense for `np.ndarray(...)` not to work because you can construct an array in a thousand ways. Our arrays are only thin wrappers around tensors, so we don't have that issue.
If you want to be extra sure that it's being used as expected, you can put an assert t is None or ìsinstance(t, torch.Tensor) in the `__init__` constructor.

ev-br:
OK, thanks. Having no public API does make it easier. 
FTR, I still believe constructing a thing from a tensor is a prime example for a `from_tensor` alternative constructor. That said, this is not my hill not today anyway, and the last commit ditches the `get()` getter, adds `ndarray.__init__(self, t: torch.Tensor)`, and makes `ndarray.tensor` public.

lezcano (approved):
Nice clean-up!

PR #82, opened 2023-03-11
MAINT: split remaining functions into normalizations and implementations

ev-br:
This finishes up the move of all work to be done on the torch level in `detail`, and the wrappers are as thin as it gets AFAICS.

### DIFF
--- a/torch_np/_detail/_reductions.py
+++ b/torch_np/_detail/_reductions.py
@@ -317,55 +317,51 @@ def average(a, axis, weights, returned=False, keepdims=False):
     return result, wsum
 
 
-def average_noweights(a_tensor, axis, keepdims=False):
-    result = mean(a_tensor, axis=axis, keepdims=keepdims)
-    scl = torch.as_tensor(a_tensor.numel() / result.numel(), dtype=result.dtype)
+def average_noweights(a, axis, keepdims=False):
+    result = mean(a, axis=axis, keepdims=keepdims)
+    scl = torch.as_tensor(a.numel() / result.numel(), dtype=result.dtype)
     return result, scl
 
 
-def average_weights(a_tensor, axis, w_tensor, keepdims=False):
+def average_weights(a, axis, w, keepdims=False):
 
     # dtype
     # FIXME: 1. use result_type
     #        2. actually implement multiply w/dtype
-    if not a_tensor.dtype.is_floating_point:
+    if not a.dtype.is_floating_point:
         result_dtype = torch.float64
-        a_tensor = a_tensor.to(result_dtype)
+        a = a.to(result_dtype)
 
-    result_dtype = _dtypes_impl.result_type_impl([a_tensor.dtype, w_tensor.dtype])
+    result_dtype = _dtypes_impl.result_type_impl([a.dtype, w.dtype])
 
-    a_tensor = _util.cast_if_needed(a_tensor, result_dtype)
-    w_tensor = _util.cast_if_needed(w_tensor, result_dtype)
+    a = _util.cast_if_needed(a, result_dtype)
+    w = _util.cast_if_needed(w, result_dtype)
 
     # axis=None ravels, so store the originals to reuse with keepdims=True below
-    ax, ndim = axis, a_tensor.ndim
+    ax, ndim = axis, a.ndim
 
     # axis
     if axis is None:
-        (a_tensor, w_tensor), axis = _util.axis_none_ravel(
-            a_tensor, w_tensor, axis=axis
-        )
+        (a, w), axis = _util.axis_none_ravel(a, w, axis=axis)
 
     # axis & weights
-    if a_tensor.shape != w_tensor.shape:
+    if a.shape != w.shape:
         if axis is None:
             raise TypeError(
                 "Axis must be specified when shapes of a and weights " "differ."
             )
-        if w_tensor.ndim != 1:
+        if w.ndim != 1:
             raise TypeError("1D weights expected when shapes of a and weights differ.")
-        if w_tensor.shape[0] != a_tensor.shape[axis]:
+        if w.shape[0] != a.shape[axis]:
             raise ValueError("Length of weights not compatible with specified axis.")
 
         # setup weight to broadcast along axis
-        w_tensor = torch.broadcast_to(
-            w_tensor, (a_tensor.ndim - 1) * (1,) + w_tensor.shape
-        )
-        w_tensor = w_tensor.swapaxes(-1, axis)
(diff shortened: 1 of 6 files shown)


torch_np/_wrapper.py
```diff
@@ -855,8 +784,10 @@ def i0(x: ArrayLike):
 def isscalar(a):
     # XXX: this is a stub
     try:
-        arr = asarray(a)
-        return arr.size == 1
+        from ._ndarray import asarray
+
+        t = asarray(a).get()
+        return t.numel() == 1
```
lezcano, line 790: This doesn't feel right. Even if this function is trivial, it should have a split like all the others.

torch_np/_detail/_reductions.py
```diff
@@ -404,3 +421,23 @@ def quantile(a_tensor, q_tensor, axis, method, keepdims=False):
     if keepdims:
         result = _util.apply_keepdims(result, ax, ndim)
     return result
+
+
+def percentile(
+    a_tensor,
+    q_tensor,
```
lezcano, line 428: Everything in this file is a tensor, so no need to repeat ourselves. You can call these `a, q` and be less verbose.

torch_np/_detail/implementations.py
```diff
@@ -450,6 +462,10 @@ def indices(dimensions, dtype=int, sparse=False):
 
 
 def bincount(x_tensor, /, weights_tensor=None, minlength=0):
+    if x_tensor.numel() == 0:
+        # edge case allowed by numpy
+        x_tensor = torch.as_tensor([], dtype=int)
```
lezcano, line 467: `x = x.new_empty(()) `may do? This way if `x_tensor` was of an incorrect dtype we'll error out as well.

lezcano (approved):
A few minor points, otherwise it LGTM

torch_np/_detail/implementations.py
ev-br: Don't think so: 

```
(Pdb) x.new_empty((), dtype=int)
tensor(140087024053344)
```

has `.numel() == 1` instead of zero.

torch_np/_wrapper.py
ev-br: OK, done in https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/82/commits/248c02bc8ba2a237471f3b096038642a0101514e

To me, it feels squarely in the cure being worse than a disease territory but hey.

torch_np/_detail/implementations.py
lezcano: Sorry, I meant `x = x.new_empty(0)`

torch_np/_wrapper.py
lezcano: So, to put this a bit more in context, making everything uniform in a strict way, even for simple ops like this comes from the following plan:
- Split everything into preprocessing and actual implementation
- Automatise as much of the preprocessing as possible via decorators or any other method. This induces a split API / impl.
- Have a look at how many exceptions do we have left after this. If there's not that many, then that means that we were able to automatise the preprocessing well-enough and the API / impl does not make sense anymore. Then, impl can be moved to the API, again, and have the API part completely removed.

Note that his is just a 3 step program towards a large refactoring. As any refactoring, this could have been done in 1 step, but, if you do that, you may find that step 3 above does not really hold and you cannot really do the refactoring. With this plan, if 3 does not hold, we stay at 2, which is still a better state than the initial one.

torch_np/_detail/implementations.py
ev-br: Ah, right. Did not know about `new_empty` before. Thanks. 
Still needs a `new_empty(0, dtype=int)` for safe casting from the default float64 to integer; reads way better anyway. Thank you.

torch_np/_wrapper.py
ev-br: ISTM we're on the same page here on the big picture. The technical Q going forward is that the `normalizer` decorator allows optional arguments here, which it did not in gh-70. I can either move it over there or leave it here at the top of the 'stack' of gh-70 <- gh-80 <- gh-81 <-gh-82 , whichever make it easier to review.

lezcano: I'm fine leaving it here, unless you really prefer to float it to gh70.

torch_np/_detail/_reductions.py
```diff
@@ -371,36 +367,73 @@ def average_weights(a_tensor, axis, w_tensor, keepdims=False):
...
+        # NumPy documents that `overwrite_input` MAY modify inputs:
+        # https://numpy.org/doc/stable/reference/generated/numpy.percentile.html#numpy-percentile
+        # Here we choose to work out-of-place because why not.
+        pass
+
+    if interpolation is not None:
+        raise ValueError("'interpolation' argument is deprecated; use 'method' instead")
+
+    if (0 > q).any() or (q > 1).any():
+        raise ValueError("Quantiles must be in range [0, 1], got %s" % q)
```
lezcano, line 390: I missed this one. `any()` synchronises and it creates a graph break in TorchDynamo. Let's leave it for now, but we may need to remove these checks in the future.

For now, let's try to use it as little as possible, so let's prefer `((0 > q) | (q > 1)).any()`

torch_np/_detail/implementations.py
lezcano: If you don't need the dtype/device/layout of `x`, then you are probably better off calling `torch.empty`.

This reminds me that, once we try things with different dtypes, we'll probably need to be careful when using constructors, as calling `torch.empty` will create a tensor in CPU, and we probably want to maintain the device of the input tensor in our intermediary tensors. Now, this is just me thinking out loud. Nothing t do here.

torch_np/_wrapper.py
lezcano: Also, I belive you were right. After seeing the machinery necessary to accommodate this one function, I believe that the original solution was good enough. Sorry for the trouble. If you feel like it, feel free to revert.

On a different note, I believe we will want parameters to `normalizer` at some point .Let's discuss some possibilities in a week, once I'm back and I've reviewed gh70.

ev-br: Agreed, we'll need parameters anyway, and there is a solution now, can as well keep it. Let's discuss it when you're back, enjoy your vacation!

torch_np/_detail/_reductions.py
ev-br: Since this function delegates to `torch.quantile` below,  and `torch.quantile` has a similar check, I just removed this check here. The only price to pay is `assert_raises((RuntimeError, ValueError), ...)` in place of numpy's `assert_raises(ValueError, ...)` --- and we're using this in multiple places anyway.

PR #83, opened 2023-03-14
Use return annotation to wrap tensors into ndarrays/sequences of ndarrays

ev-br:
Annotate return values of wrapper functions to trigger generic postprocessing. No annotation mean no postprocessing of scalar returns; `-> NDArray` triggers `ndarray(tensor)`  etc. There's again a need for a special annotation for functions which return either an array or a sequence of arrays (`atleast_1d`, `where`, `unique`). Other than that, all looks straightforward.

on top of gh-82, so it's the top of the stack based off gh-70.

NB: postprocessing currently invokes functions from `_helpers`. Those are straighforward to clean up/inline, once the rest of the "stack" is in.


torch_np/_normalizations.py
```diff
@@ -152,7 +156,31 @@ def wrapped(*args, **kwds):
...
+            # handle returns
+            r = sig.return_annotation
+            if r == NDArray:
+                return _helpers.array_from(result)
+            elif r == inspect._empty:
+                return result
+            elif hasattr(r, "__origin__") and r.__origin__ in (list, tuple):
+                # this is tuple[NDArray] or list[NDArray]
+                # XXX: change to separate tuple and list normalizers?
+                return r.__origin__(_helpers.tuple_arrays_from(result))
```
ev-br, line 169: I'm actually not sure if this is the right thing to do. The goal here is to get a `tuple` from `tuple[NDArray]`.

torch_np/_normalizations.py
```diff
@@ -104,7 +105,46 @@ def normalize_this(arg, parm, return_on_failure=_sentinel):
...
+    if isinstance(result, (tuple, list)):
+        seq = type(result)
+        return seq(_helpers.tuple_arrays_from(result))
+    else:
+        return _helpers.array_from(result)
+
+
+postprocessors = {
+    NDArray: postprocess_ndarray,
+    OutArray: postprocess_out,
```
lezcano, line 140: As discussed, I think we should be able to implement the `out=` keyword generically. Let's catch-up on this today.

lezcano (commented):
Numpy returns either a Tensor, or a list / tuple or Arrays. Now, why do we need the annotations if this information is readily available in the type of `result` after executing the function? I think we should be able to post process everything simply based on that.

PR #84, opened 2023-03-15
ENH: add a naive divmod, un-xfail relevant tests

ev-br:
Implement `np.divmod(x1, x2)` and `ndarray.__divmod__`. These are unusual in that they allow two output arrays, so it's either complicate and obfuscate the ufunc machinery just for these two, or special case. This PR opts for the latter :-).

Since there is no pytorch-native divmod (cf https://github.com/pytorch/pytorch/issues/90820 ), this PR just does `(x1 // x2, x1 % x2)` in the wrapper level.  

This PR is on top of gh-83, for clarity.

### DIFF
--- a/torch_np/_binary_ufuncs.py
+++ b/torch_np/_binary_ufuncs.py
@@ -57,3 +57,58 @@ for name in __all__:
     decorated.__qualname__ = name  # XXX: is this really correct?
     decorated.__name__ = name
     vars()[name] = decorated
+
+
+# a stub implementation of divmod, should be improved after
+# https://github.com/pytorch/pytorch/issues/90820 is fixed in pytorch
+#
+# Implementation details: we just call two ufuncs which have been created
+# just above, for x1 // x2 and x1 % x2.
+# This means we are normalizing x1, x2 in each of the ufuncs --- note that there
+# is no @normalizer on divmod.
+
+
+def divmod(
+    x1,
+    x2,
+    /,
+    out=None,
+    *,
+    where=True,
+    casting="same_kind",
+    order="K",
+    dtype=None,
+    subok: SubokLike = False,
+    signature=None,
+    extobj=None,
+):
+    out1, out2 = None, None
+    if out is not None:
+        out1, out2 = out
+
+    kwds = dict(
+        where=where,
+        casting=casting,
+        order=order,
+        dtype=dtype,
+        subok=subok,
+        signature=signature,
+        extobj=extobj,
+    )
+
+    # NB: use local names for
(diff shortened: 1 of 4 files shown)


ev-br:
Two more ufuncs with two optional out= inputs and two OutArray outputs : ldexp and frexp. Need a dedicated result annotation?

torch_np/_binary_ufuncs.py
```diff
@@ -57,3 +57,58 @@ def wrapped(
...
+        order=order,
+        dtype=dtype,
+        subok=subok,
+        signature=signature,
+        extobj=extobj,
+    )
+
+    # NB: use local names for
+    quot = floor_divide(x1, x2, out=out1, **kwds)
+    rem = remainder(x1, x2, out=out2, **kwds)
```
lezcano, line 101: Note that here (and in general) if we want these functions to be differentiable in PyTorch, we should not use their `out=` variant. We should implement the `out=` behaviour manually.

torch_np/_normalizations.py
```diff
@@ -154,33 +189,17 @@ def wrapped(*args, **kwds):
...
-                return _helpers.result_or_out(
-                    result, out, promote_scalar=promote_scalar_out
-                )
-            else:
-                raise ValueError(f"Unknown return annotation {return_annotation}")
+            postprocess = postprocessors.get(r, None)
+            if postprocess:
+                kwds = {"promote_scalar": promote_scalar_out}
+                result = postprocess(result, **kwds)
+            return result
```
lezcano, line 202: Also dupped?

lezcano (commented):
Let's figure out what's going on with that duplicated return code, otherwise the implementations LGTM. Note that `divmod` can also be implemented via a divison and a substraction, but I'm happy with this implementation as well as it's simpler.

lezcano (approved):
I meant

torch_np/_binary_ufuncs.py
ev-br: Here these are wrapper `floor_divide` and `remainder`, so `torch.remainder` and `torch.floor_divide` never see the `out=`. 
Also NB: this will need a rework anyway, apparently there are more (out1, out2) ufuncs, ldexp and frexp. So `Tuple[NDArray]` return annotation will need to appear after we settle on the generic machinery.

ev-br:
> Let's figure out what's going on with that duplicated return code

That's just an artefact from giving optional return arguments to `normalizer`: the whole stack will need rebases now that we've reworked the `normalizer` in gh-70 which is the bottom of the stack. Will rebase when merging from bottom up.

ev-br:

Issue #86, opened 2023-03-16
Short- to medium-term planning


ev-br:
With the MVP stage being finalized in the "stack" of gh-70 -- gh-83, let's think of prioritizing future work. Small fixes aside,  larger work items at the moment are

- [ ] NumPy API coverage. https://github.com/Quansight-Labs/numpy_pytorch_interop/blob/main/autogen/numpy_api_dump.py lists numpy apis which are not implemented at all. The list needs some prioritization, too.
- [ ] fancy/advanced indexing and related functions
- [ ] ufunc dtype resolution and its interaction with the registered loops. Our current implementation does not have any of this: explicit `dtype=...` raises NotImplementedError, and internal loops are delegated to whatever pytorch is doing. 
- [ ] dtype resultion/promotion/casting in general. 
- [ ] `where = ...` arguments to ufuncs and related gather/scatter type functionality
- [ ] non-default array orders and fortran-ordered arrays (do we support this at all?)

rgommers:
Thanks for writing this up Evgeni!

>  NumPy API coverage. https://github.com/Quansight-Labs/numpy_pytorch_interop/blob/main/autogen/numpy_api_dump.py lists numpy apis which are not implemented at all. The list needs some prioritization, too.

Do you want to do this in a second issue? I guess we need the diff between the names in `numpy_api_dump.py` and `dir(numpy)` plus the same for the key submodules - `fft`, `linalg,` random`.

For the rest of the items: 
- `dtype=` sounds like the top prio to me. 
- Advanced indexing should also be high on the list I'd think - PyTorch should already have implementations that mostly match NumPy, so hopefully this isn't a ton of work.
- dtype resultion/promotion/casting in general: we probably won't get every last detail right here, and it'll change for scalars with NEP 50. I'd prefer to match the new behavior after NEP 50 becomes the default - that should be easier to do, and future-proof. For the rest, I'm not sure what falls in this bucket exactly.
- `where=` seems low-prio, it isn't used much and it's nontrivial to implement probably
- array ordering: don't support it at all I'd say.

ev-br:
> Do you want to do this in a second issue? I guess we need the diff between the names in numpy_api_dump.py and dir(numpy) plus the same for the key submodules - fft, linalg, random`.

Let's take it to gh-87 indeed.

ev-br:
[...]

PR #89, opened 2023-03-19
dtype=... arguments of ufuncs

ev-br:
This implementation is a bit simpler than numpy: we do not have the notion of ufunc loop types (np.add.types etc), so we just cast input tensors to the `result_type(dtype, out.dtype)`, and ask pytorch to do computations in that dtype.

This is a superset of what numpy is doing AFAICS: it may differ in corner cases (`np.sin(3, dtype=bool`, anyone?), mainly where numpy either raises or returns nonsense. 

Cross-ref https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/86#issuecomment-1471732287.

### DIFF
--- a/torch_np/_detail/_util.py
+++ b/torch_np/_detail/_util.py
@@ -137,7 +137,7 @@ def axis_none_ravel(*tensors, axis=None):
         return tensors, axis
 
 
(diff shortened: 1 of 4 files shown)


ev-br:

Issue #85, opened 2023-03-15
matmul with out array is broken


ev-br:
consider `np.matmul(a, b, out)` with `a.shape=(5, 2)` and `b.shape=(2, 2)`. The out array is `out.shape=(5, 2)`, but the ufunc machinery attempts to broadcast `a` and `b` against `out`, which of course fails at `broadcast(b, out)`.

So need to special-case matmul in the ufunc machinery.

ev-br:
Fixed by gh-90

PR #90, opened 2023-03-19
BUG: Fix matmul with out= arrays

ev-br:
fixes gh-85

`matmul(x1, x2, out)` does not broadcast x1, x2 against out, like other ufuncs do.

### DIFF
--- a/torch_np/tests/numpy_tests/core/test_multiarray.py
+++ b/torch_np/tests/numpy_tests/core/test_multiarray.py
@@ -2605,7 +2605,6 @@ class TestMethods:
         if HAS_REFCOUNT:
             assert_(sys.getrefcount(a) < 50)
 
-    @pytest.mark.xfail(reason="TODO: implement np.dot")
     def test_size_zero_memleak(self):
         # Regression test for issue 9615
         # Exercises a special-case code path for dot products of length
@@ -6072,11 +6071,11 @@ class TestMatmul(MatmulCommon):
         assert not np.any(c)
 
 
-@pytest.mark.xfail(reason='TODO: @')
 class TestMatmulOperator(MatmulCommon):
     import operator
     matmul = operator.matmul
 
+    @pytest.mark.skip(reason="no __array_priority__")
     def test_array_priority_override(self):
 
         class A:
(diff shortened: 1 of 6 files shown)


torch_np/tests/numpy_tests/core/test_multiarray.py
```diff
@@ -5919,9 +5925,10 @@ def test_matrix_matrix_values(self):
         assert_equal(res, tgt12_21)
 
 
-@pytest.mark.xfail(reason='TODO: matmul (ufunc wrapping goes south?)')
 class TestMatmul(MatmulCommon):
-    matmul = np.matmul
+
+    def setup_method(self):
+        self.matmul = np.matmul
```
ev-br, line 5931: This change is a bit curious. Storing ufuncs as class attributes seems to work ok in scripts/interactive interpreter, but breaks down in combination with pytest:

the argument processing/normalization machinery picks up the extra `self`-like argument, which leads to 

```
>           tensor = torch.as_tensor(obj)
E           RuntimeError: Could not infer dtype of TestMatmul
```

so there's some spooky action on a distance between meta-stuff pytest is doing and what we do here.

torch_np/tests/numpy_tests/core/test_multiarray.py
```diff
@@ -5974,7 +5981,7 @@ def test_out_contiguous(self):
         # test out non-contiguous
         out = np.ones((5, 2, 2), dtype=float)
         c = self.matmul(a, b, out=out[..., 0])
-        assert c.base is out
+        assert c._tensor._base is out._tensor     # FIXME: self.tensor (no underscore)
```
ev-br, line 5984: This will get cleaned up before merging, after the base of the stack gets in.

lezcano (commented):
So, I think the broadcasting issue for matmul will simply go away when we implement the `out=` semantics by doing the broadcasting wrt. the out parameter after the computation rather than before. The other matmul fixes from this PR LGTM, so what about simply merging those and revisiting the `out=` behaviour later?

ev-br:
> The other matmul fixes from this PR LGTM, so what about simply merging those and revisiting the out= behaviour later?

Works for me :-). Merging this PR is blocked by the normalizations PR though, as this one is at the top of the stack.

ev-br:

PR #93, opened 2023-03-24
wrap returns out annotation

ev-br:
<s>Use a return annotation for functions with an `out=` argument. Other returns are wrapped generically.</s>

Detect the out argument in signatures and implement the out= semantics. 

supersedes and closes gh-83


torch_np/_normalizations.py
```diff
@@ -144,7 +179,12 @@ def wrapped(*args, **kwds):
                 for name, arg in kwds.items()
             }
             result = func(*args, **kwds)
-            result = wrap_tensors(result)
+
+            out = None
+            if sig.return_annotation == OutArray:
+                result, out = result
+
+            result = wrap_tensors(result, out, promote_scalar_result)
```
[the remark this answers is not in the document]
ev-br: I'm going to prepare a version without return annotations.

ev-br:
Force pushed a version without type annotations, almost identical to https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/93#discussion_r1149073645. 
I was wanting to avoid dispatching on argument names ("out"), but the alternatives seem clunky, so let's roll with this and see if want to improve at leisure.

One possibly semi-missing bit is divmod (again!), which has several ways of accepting the out arrays. Working out all details of what numpy _really_ does requires taking over `numpy/core/tests/test_umath.py`. This is rather large and rather hairy, so best not conflate it with the cleanup that is this PR.

torch_np/_binary_ufuncs.py
```diff
@@ -77,69 +77,57 @@ def matmul(
...
 def divmod(
-    x1,
-    x2,
+    x1: ArrayLike,
+    x2: ArrayLike,
+    out1: Optional[NDArray] = None,
+    out2: Optional[NDArray] = None,
     /,
-    out=None,
+    out: Optional[tuple[NDArray]] = (None, None),
```
lezcano, line 89: nit `tuple[Optional[NDArray], Optional[NDArray]]`.

torch_np/_binary_ufuncs.py
```diff
@@ -77,69 +77,57 @@ def matmul(
...
-    kwds = dict(
-        where=where,
-        casting=casting,
-        order=order,
-        dtype=dtype,
-        subok=subok,
-        signature=signature,
-        extobj=extobj,
+    tensors = _helpers.ufunc_preprocess(
+        (x1, x2), out, True, casting, order, dtype, subok, signature, extobj
```
lezcano, line 110: As discussed `ufunc_preprocess` should not take an `out` parameter as setting an `out=` parameter should not affect the type of the computation. Or is this fixed in some other PR and we'll just need to update this one afterwards?

lezcano (approved):
I agree that dispatching on name is not the best. Another way of doing this would be to have an annotation `Output = Union[NDArray, Tuple[NDArray...], Tuple[Any...]]` (for example, it's not like we're using them for typecheking really) and then dispatch on that. Either works for me.

About `divmod`... my my. Yeah, your one-off solution LGTM.

torch_np/_binary_ufuncs.py
ev-br: Why? It's either `(None, None)` or `(NDArray, NDArray)`; can't have `(out1, None)` (I _think_)

torch_np/_binary_ufuncs.py
ev-br: It's still in the TODO pipeline. Will do after the current "stack" is done.

ev-br:
Thanks Mario

torch_np/_binary_ufuncs.py
lezcano: `tuple`, as opposed, to `list` takes the number of arguments that it accepts explicitly. As such, there you are declaring there a tuple of length 1 and one NDArray. 
Then, `Optional[tuple[NDArray, NDArray]]` accepts either a tuple of ndarrays (fine) or `None` (not `(None, None)`).
What you want is to have a (partial) function that accepts `tuple[Optional[NDArray], Optional[NDArray]]` and errors out when just one of the elements of the tuple is `None`.

lezcano: Feel free to patch this small correction in any other PR.

ev-br: Ouch. python typing meta-language is seriously weird.

lezcano: Note that this is not specific to Python, the same would be true in C++. This would be an `std::tuple<std::optional<array, std::optional<array>>` rather than a `std::optional<std::tuple<array, array>>` :)

PR #94, opened 2023-03-25
Merge wrapper and implementations

ev-br:
Undo the split between decorated functions and torch implementations: the formers are nothing but fall-throught to tha latters now that all argument pre/postprocessing is contained in normalization decorators.
While at it, merge `_funcs.py` and `_wrapper.py`, the split was purely technical.
The diff looks scary, but this is mainly just shuffling things around and trimming a small amount of dead weight.


torch_np/_ndarray.py
```diff
@@ -358,8 +357,7 @@ def reshape(self, *shape, order="C"):
 
     def sort(self, axis=-1, kind=None, order=None):
         # ndarray.sort works in-place
```
lezcano, line 359: If so, do we need to do a `copy_`?

torch_np/_funcs.py
```diff
@@ -188,60 +1380,63 @@ def rollaxis(a: ArrayLike, axis, start=0):
...
     return result
 
 
 # leading underscore since arr.flatten exists but np.flatten does not
 @normalizer
 def _flatten(a: ArrayLike, order="C"):
-    result = _impl._flatten(a)
+    if order != "C":
+        raise NotImplementedError
+    # return a copy
```
lezcano, line 1432: Nit. It **may** return a copy

torch_np/_funcs.py
```diff
@@ -503,3 +1698,203 @@ def median(
...
+def i0(x: ArrayLike):
+    result = torch.special.i0(x)
+    return result
+
+
+@normalizer(return_on_failure=False)
+def isscalar(a: ArrayLike):
+    # XXX: this is a stub
+    if a is False:
+        return a
```
lezcano, line 1872: What does this construction mean?

torch_np/_funcs.py
```diff
@@ -503,3 +1698,203 @@ def median(
...
+    result = torch.zeros_like(x, dtype=torch.bool)
+    if result.ndim == 0:
+        result = result.item()
+    return result
+
+
+@normalizer
+def isreal(x: ArrayLike):
+    if torch.is_complex(x):
+        return torch.as_tensor(x).imag == 0
```
lezcano, line 1831: `x` is already a tensor.

torch_np/_funcs.py
```diff
@@ -503,3 +1698,203 @@ def median(
...
+        tol = tol * torch.finfo(a.dtype).eps
+
+    mask = torch.abs(a.imag) < tol
+    return a.real if mask.all() else a
+
+
+@normalizer
+def iscomplex(x: ArrayLike):
+    if torch.is_complex(x):
+        return torch.as_tensor(x).imag != 0
```
lezcano, line 1821: `x` is a tensor.

lezcano (commented):
The merge LGTM. We are still missing to merge the other files within `_detail`, but this first step looks good.

I took this chance to go through the implementations here and flag a few issues with them. Just minor things all throughout.

Style Nit. Very often you name a variable result, just to return it in the next line. I think it's cleaner and more concise to simply do `return torch.fn`.

There's a number of places where we use methods rather than their `torch.fn` versions. It'd be good to be consistent all throughout. I flagged a few of these, but there's more (roll, squeeze, reshape...)

torch_np/_wrapper.py
```diff
@@ -1,862 +0,0 @@
...
-# 1. Mapping of the numpy layout ('C', 'K' etc) to torch layout / memory_format.
-# 2. np.dtype <-> torch.dtype
-# 3. numpy type casting rules (to be cleaned up in numpy: follow old or new)
-#
-# 4. wrap/unwrap/wrap patterns:
-#   - inputs are scalars, output is an array
-#   - two-arg functions (second may be None)
-#   - first arg is a sequence/tuple (_stack familty, concatenate, atleast_Nd etc)
-#   - optional out arg
-
```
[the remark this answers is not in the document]
ev-br: It's indeed removed together with out outdated comments in a PR up the stack.

torch_np/_funcs.py
```diff
@@ -1,21 +1,1043 @@
...
+from torch import broadcast_shapes
+
+
+@normalizer
+def broadcast_arrays(*args: ArrayLike, subok: SubokLike = False):
+    res = torch.broadcast_tensors(*args)
+    return res
+
+
+# FIXME: this needs fixing
```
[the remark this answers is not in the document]
ev-br: This whole function is untested an needs fixing.

torch_np/_funcs.py
```diff
@@ -1,21 +1,1043 @@
...
+    if equal_nan:
+        nan_loc = (torch.isnan(a1_t) == torch.isnan(a2_t)).all()
+        if nan_loc:
+            # check the values
+            result = a1_t[~torch.isnan(a1_t)] == a2_t[~torch.isnan(a2_t)]
+        else:
+            return False
+    else:
+        result = a1_t == a2_t
+    return bool(result.all())
```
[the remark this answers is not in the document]
ev-br: It is:
````
(Pdb) torch.equal(torch.as_tensor(torch.nan), torch.as_tensor(torch.nan), equal_nan=False)
*** TypeError: equal() got an unexpected keyword argument 'equal_nan'
```

torch_np/_funcs.py
```diff
@@ -1,21 +1,1043 @@
...
+def mintypecode():
+    raise NotImplementedError
+
+
+def nan_to_num():
+    raise NotImplementedError
+
+
+def asfarray():
+    raise NotImplementedError
```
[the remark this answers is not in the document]
ev-br: These are just stubs to minimize the churn in tests of higher prio functions. These only need to import.

torch_np/_funcs.py
```diff
@@ -1,21 +1,1043 @@
...
+def common_type():
+    raise NotImplementedError
+
+
+def mintypecode():
+    raise NotImplementedError
+
+
+def nan_to_num():
+    raise NotImplementedError
```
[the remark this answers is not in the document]
ev-br: Once it floats to the top of the prio queue, sure.

torch_np/_funcs.py
ev-br: `if a is False: return False`?

torch_np/_funcs.py
```diff
@@ -1,21 +1,1043 @@
...
+    dtype = _dtypes_impl.result_type_impl((a.dtype, b.dtype))
+    a = _util.cast_if_needed(a, dtype)
+    b = _util.cast_if_needed(b, dtype)
+    result = torch.isclose(a, b, rtol=rtol, atol=atol, equal_nan=equal_nan)
+    return result
+
+
+def allclose(a: ArrayLike, b: ArrayLike, rtol=1e-05, atol=1e-08, equal_nan=False):
+    result = isclose(a, b, rtol, atol, equal_nan=equal_nan)
+    return result.all()
```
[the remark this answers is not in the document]
ev-br: This duplicates the `cast_if_needed` logic from `isclose` but OK, done.

torch_np/_funcs.py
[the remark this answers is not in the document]
lezcano: Ok, what about
```suggestion
def _tensor_equal(a1, a2, equal_nan=False):
    # Implementation of array_equal/array_equiv.
    if equal_nan:
      return (a1.shape == a2.shape) and ((a1 == a2) | (torch.isnan(a1) & torch.isnan(a2))).all().item()
    else:
      return torch.equal(a1, a2)
```

torch_np/_ndarray.py
lezcano: We have to perform the copy in-place to preserve the semantics:
```python
>>> import numpy as np
>>> a = np.array([1, 3, 2])
>>> id(a)
140581065530512
>>> a.data
<memory at 0x7fdb94820dc0>
>>> a.sort()
>>> a
array([1, 2, 3])
>>> id(a)
140581065530512
>>> a.data
<memory at 0x7fdb94820dc0>
```

torch_np/_funcs.py
lezcano: Yeah, but `a` is an `ArrrayLike`, when is it going to be the literal `False`?

ev-br: OK, fair enough. Thanks for spotting this. Removed.

torch_np/_ndarray.py
ev-br: Indeed. `self.tensor.copy_(_funcs._sort)` it is then.

torch_np/_funcs.py
lezcano: This uncovered a latent bug:
```
    @normalizer(return_on_failure=False)
    def isscalar(a: ArrayLike):
        # XXX: this is a stub
>       return a.numel() == 1
E       AttributeError: 'bool' object has no attribute 'numel'
```
It may be better to just implement this function without the normalizer and remove the `return_on_failure` flag, as you initially had?

ev-br: All right. Here's what's going on: `isscalar(np.int)`  (`np.int`is not a scalar!) fails to convert its argument to tensor, and the `return_on_failure` code path feeds `False` into the the body of the function. This is in fact requested by the normalizer argument three lines above, via `normalizer(return_on_failure=False)`.  So, can either 
- put back the `if a is False: return False`  clause + a code comment to not confuse future readers
- rip the `return_on_failure` code path and normalize manually (and possibly will need to normalize manually other functions from the not-yet-implemented numpy apis, who knows)

Your preference @lezcano ?

torch_np/_funcs.py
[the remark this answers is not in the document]
ev-br: done

torch_np/_funcs.py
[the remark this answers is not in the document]
ev-br: Took a quick look. Pretty typical, this needs several workarounds around torch limitations (== numpy being too lax, depending on the POV), so let's kick it down the road for now.

torch_np/_funcs.py
lezcano: Probably the simplest option is the second one, as this function can be implemented as
```python
def isscalar(a):
    try:
        return asarray(a).size == 1
    except:
        return False
```
and it's very much one of a kind.

torch_np/_funcs.py
```diff
@@ -1,21 +1,1043 @@
...
+        out = []
+        for dim in reversed(shape):
+            out.append(index % dim)
+            index = index // dim
+        result.append(tuple(reversed(out)))
+    return result
+
+
+def ravel_multi_index(multi_index, dims, mode="raise", order="C"):
+    return sum(idx * dim for idx, dim in zip(multi_index, dims))
```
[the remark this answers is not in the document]
lezcano: We use the normalization approach (among other things) to handle the fact that NumPy often accepts anythng thats smells like an array in place of an array (e.g. a list of lists) and then returns an array. This function would deviate from that behaviour because we are not normalizing it. In general, whenever we can, we should use the normalizer to avoid these sly deviations from the NumPy semantics.

lezcano (approved):
Missing two minor things. Approving this contingent on fixing those two.

This one got us -600LOC, not too bad!

torch_np/_funcs.py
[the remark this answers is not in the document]
ev-br: OK, let's remove it for now and add it back to the TODO priority queue.

ev-br:
CI is green, the two last comments are addressed, merging.

Issue #87, opened 2023-03-16
NumPy API surface : plan/prioritize the coverage


ev-br:
Splitting it off gh-86, here's the difference in API surfaces of NumPy and this wrapper. We can edit the order  to reflect priorities:

```
>>> import numpy as np
>>> import torch_np as tnp
>>> npset = set(x for x in dir(np) if not x.startswith('_') and not inspect.ismodule(x) and not x[0].isupper())
>>> tnpset = set(dir(tnp))
>>> for name in sorted(npset - tnpset):
...        print("-[ ]", name)
```

-[ ] add_docstring
-[ ] add_newdoc
-[ ] add_newdoc_ufunc
-[ ] append
-[ ] apply_along_axis
-[ ] apply_over_axes
-[ ] argpartition
-[ ] array2string
-[ ] array_repr
-[ ] array_str
-[ ] asanyarray
-[ ] asarray_chkfinite
-[ ] ascontiguousarray
-[ ] asfortranarray
-[ ] asmatrix
-[ ] bartlett
-[ ] base_repr
-[ ] binary_repr
-[ ] bitwise_not
-[ ] blackman
-[ ] block
-[ ] bmat
-[ ] broadcast
-[ ] busday_count
-[ ] busday_offset
-[ ] busdaycalendar
-[ ] byte_bounds
-[ ] bytes_
-[ ] c_
-[ ] cast
-[ ] cfloat
-[ ] char
-[ ] character
-[ ] chararray
-[ ] choose
-[ ] clongdouble
-[ ] clongfloat
-[ ] compare_chararrays
-[ ] compat
-[ ] complex256
-[ ] complex_
-[ ] compress
-[ ] convolve
-[ ] copyto
-[ ] correlate
-[ ] cross
-[ ] ctypeslib
-[ ] datetime64
-[ ] datetime_as_string
-[ ] datetime_data
-[ ] delete
-[ ] deprecate
-[ ] deprecate_with_doc
-[ ] digitize
-[ ] disp
-[ ] e
-[ ] ediff1d
-[ ] einsum
-[ ] einsum_path
-[ ] emath
-[ ] euler_gamma
-[ ] extract
-[ ] fastCopyAndTranspose
-[ ] fft
-[ ] find_common_type
-[ ] fix
-[ ] flatiter
-[ ] flexible
-[ ] float128
-[ ] format_float_positional
-[ ] format_float_scientific
-[ ] format_parser
-[ ] frexp
[...]

PR #95, opened 2023-03-25
Add filter windows, tensordot

ev-br:
cross-ref https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/87#issuecomment-1478036400


lezcano (commented):
The docs say that `M` is an int, and so do the examples in the docs. As such, it is the tests that are wrong. We can still make the test pass implementing this undocumented behaviour, or simply remove them. At any rate, `M` should be of type `int` or at most `Union[int, ArrayLike]`. What I'm trying to avoid here is that unconditional wrapping of the `int` and an unconditional call to `item()` (which is data dependent and would cause issues when tracing through it).

ev-br:
The issue here is that an int is an array_like. And this is not the only undocumented behavior seen in the numpy test suite. So far I tried to minimize tweaking  numpy tests. So, let's remove these tests?

lezcano:
Removing the tests SGTM

lezcano:
Either that, or modify them to use an int, either is fine really.

lezcano (approved):

PR #96, opened 2023-03-28
`take()` implementation

honno:
Rudimentary `take()` implementations. Passes my Array API `test_take` at least (https://github.com/data-apis/array-api-tests/pull/173, no means comprehensive for NumPy-propers semantics of `take()`), and un-xfailing existing tests didn't uncover anything—I imagine however importing more tests we'll find a bug or two for sure though.

It's kinda awkward to implement `ndarray.take()` given the redundancies of handling the kwargs in `_wrapper.take()` too (which cannot be used due to circular import), so left it not implemented...

Hopefully this is how I should be implementing things now? I really like `@normalizer`!

PR #98, opened 2023-03-31
take() implementation, rebase

ev-br:
Rebase gh-96 on main now that all the normalizations are in and the split between the wrapper and implementations is (mostly) gone.


honno (approved):
LGTM! (nice work on the out stuff in #93)

Issue #73, opened 2023-02-28
Unsupported NumPy features and other differences w.r.t. NumPy


ev-br:
- We only aim to support numeric dtypes, which are undestood by pytorch. This rules out
    -  object arrays
    - datetimes
    - strings, chars and void dtypes
    - structured dtypes and recarrays
    - `np.longdouble` and `np.clongdouble` (a.k.a. `np.float128` and `np.complex256`, respectively)

- ndrray subclasses are out of scope.

- masked arrays are out of scope

- numpy polynomials are out of scope, both `np.poly1d`, `np.polynomial`

- `__array_wrap__` protocol is out of scope. This way, non-default `like=...`  arguments raise.

- `__array_interface__` is out of scope

- `ndarray.ctypes` attribute not supported

- In numpy, `np.flip(a)` creates a view of `a`. We follow `torch.flip` and create a copy.

- slicing with negative step is not supported: `np.arange(5)[::-1]` fails.

-------------------------------

These differences exist currently, but may be fixed if desired:

- We do not distinguish between 0-dim arrays and scalars. That is, `tnp.float32(3)` creates a zero-dim array

- We use a simplified type promotion rules for mixed-type operations: since there are no scalars, the numpy 'scalar do not upcast arrays' rule does not apply. 

- We do not implement value-based casting.

rgommers:
> `__array_wrap__` protocol is out of scope. This way, non-default `like=...`  arguments raise.

This is something that `torch.Tensor` itself uses - should that really be out of scope? It's unrelated to `like=` arguments AFAIK, that'd be `__array_function__`.

ev-br:
Thanks, edited! Have to admit, I'm not up-to-date on various `__array_dunder__` protocols. Not sure which of them does what, thus do not have a sensible suggestion on which ones we want here.

rgommers:
Is there any test case like:
```
import torch
import numpy as np
from numpy.testing import assert_raises

t1 = torch.arange(3)
t2 = torch.ones(3)
t3 = np.add(t1, t2)
assert isinstance(t3, torch.Tensor)

# Test mixing torch and numpy tensors
x1 = np.full((3,), 2)
assert isinstance(x1 + t1, torch.Tensor)  # calls numpy, which is happy to accept tensors
assert_raises(TypeError, lambda f: x1 + t1)  # calls torch, which doesn't accept numpy arrays
```

[...]

PR #100, opened 2023-04-01
add torch_np.linalg module

ev-br:
Implement linalg mappings. This PR does nearly all of linalg (sans einsum). So far this is not completely consistent on
- the type of exceptions: torch raises RuntimeErrors, sometimes `torch._C._LinAlgError`s; we mostly pass the former through and convert the latter
- zero-size arrays. NumPy handles them, but torch doesn't seem to. So just xfail the corresponding tests.

### DIFF
--- a/torch_np/linalg.py
+++ b/torch_np/linalg.py
@@ -43,7 +43,7 @@ def matrix_power(a: ArrayLike, n):
 
 
 @normalizer
-def multi_dot(inputs : Sequence[ArrayLike], *, out=None):
+def multi_dot(inputs: Sequence[ArrayLike], *, out=None):
     return torch.linalg.multi_dot(inputs)
 
 
--- a/torch_np/linalg.py
+++ b/torch_np/linalg.py
@@ -1,3 +1,5 @@
+from typing import Sequence
+
 import torch
 
 from ._detail import _dtypes_impl, _util
@@ -41,7 +43,7 @@ def matrix_power(a: ArrayLike, n):
 
 
 @normalizer
-def multi_dot(inputs, *, out=None):
+def multi_dot(inputs : Sequence[ArrayLike], *, out=None):
     return torch.linalg.multi_dot(inputs)
 
 
@@ -64,7 +66,11 @@ def lstsq(a: ArrayLike, b: ArrayLike, rcond=None):
 @normalizer
 def inv(a: ArrayLike):
     a = _atleast_float_1(a)
-    return torch.linalg.inv(a)
+    try:
+        result = torch.linalg.inv(a)
+    except torch._C._LinAlgError as e:
+        raise LinAlgError(*e.args)
+    return result
 
 
 @normalizer
--- a/torch_np/linalg.py
+++ b/torch_np/linalg.py
@@ -1,0 +1,176 @@
+import torch
+
+from ._normalizations import ArrayLike, normalizer
+from ._detail import _dtypes_impl
+from ._detail import _util
+
+
+def _atleast_float_1(a):
+    if not (a.dtype.is_floating_point or a.dtype.is_complex):
+        a = a.to(_dtypes_impl.default_float_dtype)
+    return a
+
+
+def _atleast_float_2(a, b):
+    dtyp = _dtypes_impl._result_type_impl((a.dtype, b.dtype))    
+    if not (dtyp.is_floating_point or dtype.is_complex):
+        dtyp = _dtypes_impl.default_float_dtype
+
+    a = _util.cast_if_needed(a, dtyp)
+    b = _util.cast_if_needed(b, dtyp)
+    return a, b
+
+
+
+# ### Matrix and vector products ###
+
+@normalizer
+def matrix_power(a: ArrayLike, n):
+    a = _atleat_float_1(a)
+    return torch.linalg.matrix_power(a, n)
+
+
+@normalizer
+def multi_dot(inputs, *, out=None):
+    return torch.linalg.multi_dot(inputs)    
+
+
+# ### Solving equations and inverting matrices ###
+
+@normalizer
+def solve(a: ArrayLike, b: ArrayLike):
+    a, b = _atleast_float_2(a, b)
+    return torch.linalg.solve(a, b)
+
+
+@normalizer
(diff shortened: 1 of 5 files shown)


torch_np/linalg.py
```diff
@@ -0,0 +1,224 @@
...
+    # check if empty
+    # cf: https://github.com/numpy/numpy/blob/v1.24.0/numpy/linalg/linalg.py#L1744
+    if x.numel() == 0 and _prod(x.shape[-2:]) == 0:
+        raise LinAlgError("cond is not defined on empty arrays")
+
+    result = torch.linalg.cond(x, p=p)
+
+    # Convert nans to infs unless the original array had nan entries
+    # XXX: NumPy does this: https://github.com/numpy/numpy/blob/v1.24.0/numpy/linalg/linalg.py#L1744
+    # Do we want it? .any() synchronizes
```
ev-br, line 122: Numpy goes to some lengths to convert nans to infs in `linalg.cond` --- I _think_ this is to allow making checks like `cond(a) > 1e15` without checking for nans. 
It's not free however (see below).
Do we want/need this @lezcano ?

torch_np/linalg.py
```diff
@@ -0,0 +1,224 @@
...
+
+    return result
+
+
+@normalizer
+def matrix_rank(a: ArrayLike, tol=None, hermitian=False):
+    a = _atleast_float_1(a)
+
+    if a.ndim < 2:
+        return int(not (a == 0).all())
```
lezcano, line 141: Nit. more explicit: `int((a != 0).any())`

torch_np/linalg.py
```diff
@@ -0,0 +1,224 @@
...
+        atol, rtol = tol, 0
+    return torch.linalg.matrix_rank(a, atol=atol, rtol=rtol, hermitian=hermitian)
+
+
+@normalizer
+def norm(x: ArrayLike, ord=None, axis=None, keepdims=False):
+    x = _atleast_float_1(x)
+    result = torch.linalg.norm(x, ord=ord, dim=axis)
+    if keepdims:
+        result = _util.apply_keepdims(result, axis, x.ndim)
```
lezcano, line 157: PyTorch has a `keepdim=` kwarg (note the missing s). Are the semantics not the same?

torch_np/linalg.py
```diff
@@ -0,0 +1,224 @@
...
+    return result
+
+
+@normalizer
+def svd(a: ArrayLike, full_matrices=True, compute_uv=True, hermitian=False):
+    a = _atleast_float_1(a)
+    # NB: ignore the hermitian= argument (no pytorch equivalent)
+    result = torch.linalg.svd(a, full_matrices=full_matrices)
+    if not compute_uv:
+        result = result.S
```
lezcano, line 186: Use `linalg.svdvals` (it's the way we expose this kwarg, rather than having a function that's variadic on the output.

torch_np/linalg.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+@normalizer
+def inv(a: ArrayLike):
+    a = _atleast_float_1(a)
+    try:
+        result = torch.linalg.inv(a)
+    except torch._C._LinAlgError as e:
+        raise LinAlgError(*e.args)
+    return result
```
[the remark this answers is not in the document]
lezcano: If we go down this route, it'll be a great way to also implement the `_atelast...` behaviour generically looking at the number of inputs of the function that happen to be tensors. 

We could also add support for 0-dim inputs generically this way.

lezcano (approved):

lezcano (commented):
Sorry, this looks good, but I didn't mean to approve it. Let's wait until the (minor) points are addressed and let's approve it then.

torch_np/linalg.py
ev-br: Does all(...) synchronize and cause a graph break?

torch_np/linalg.py
ev-br: This here is consistent with the rest of torch_np, where we do not use pytorch'es keepdim. IIRC it's that pytorch is not completely consistent and it's easier to emulate it generically on the wrapper level.

torch_np/linalg.py
lezcano: Yes, but so does any. Now, there's no way around it, as in this function we return a python object by looking at the data. This was a cosmetic suggestion to help being more explicit about what we want to do and avoid unnecessary negations.

torch_np/linalg.py
lezcano: I think this should generate the same code as calling `keepdim`, so that should be alright.

torch_np/linalg.py
```diff
@@ -0,0 +1,224 @@
...
+def solve(a: ArrayLike, b: ArrayLike):
+    a, b = _atleast_float_2(a, b)
+    return torch.linalg.solve(a, b)
+
+
+@normalizer
+def lstsq(a: ArrayLike, b: ArrayLike, rcond=None):
+    a, b = _atleast_float_2(a, b)
+    # NumPy is using gelsd: https://github.com/numpy/numpy/blob/v1.24.0/numpy/linalg/umath_linalg.cpp#L3991
+    return torch.linalg.lstsq(a, b, rcond=rcond, driver="gelsd")
```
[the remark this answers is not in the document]
ev-br: What is the recommended driver on CUDA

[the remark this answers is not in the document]
lezcano: https://pytorch.org/docs/stable/generated/torch.linalg.lstsq.html
>For CUDA input, the only valid driver is ‘gels’, which assumes that A is full-rank.

torch_np/linalg.py
ev-br: Yeah, `compute_uv` here looks lika a poster child illustration of a message along the lines of "what looks reasonable is time dependent" :-).

lezcano (approved):
Cool! The 0-dim and the generic implementation can be done (or not, it's fairly low prio) in a different PR.

ev-br:
> The 0-dim and the generic implementation can be done (or not, it's fairly low prio) in a different PR.

Added a line to the list at https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/73#issue-1602739140

PR #101, opened 2023-04-03
add several ndarray method/properties

ev-br:


torch_np/_ndarray.py
```diff
@@ -157,7 +166,8 @@ def view(self, dtype):
         tview = self.tensor.view(torch_dtype)
         return ndarray(tview)
 
-    def fill(self, value):
+    @normalizer
+    def fill(self, value: ArrayLike):
         self.tensor.fill_(value)
```
lezcano, line 171: In PyTorch `fill_` just works with Python scalars, it doesn't accept ArrayLikes. For ArrayLikes you need to use `copy_` and perhaps perform some shape check?

torch_np/_ndarray.py
```diff
@@ -395,6 +405,15 @@ def sort(self, axis=-1, kind=None, order=None):
...
     ### indexing ###
+    def item(self, *args):
+        if args == ():
+            return self.tensor.item()
+        elif len(args) == 1:
+            # int argument
+            return self.ravel()[args[0]]
+        else:
+            return self.__getitem__(args)
+
```
lezcano, line 416: Do we have tests for this?

torch_np/_ndarray.py
lezcano: Ah, I see. Since I saw the `ArrayLike` annotation, I tought it referred to arbitrary input arrays.

ev-br: It could have been clearer to not have the annotation indeed, but numpy allows numpy scalars and 0d arrays, so here we are.

torch_np/_ndarray.py
ev-br: Not at the moment. Don't see numpy tests in numpy proper either. Behavior with arrays seems legacy TBH, so I'd prefer to put it to the priority queue.

lezcano: If you pulled this from the NumPy implementation, let's just add a link to it and roll with it.

torch_np/_ndarray.py
lezcano: Yep. Let's just add a short comment, but otherwise LGTM.

lezcano (approved):

PR #106, opened 2023-04-05
add copyto, resize, histogram

ev-br:
on top of gh-105

### DIFF
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -6,7 +6,7 @@ pytorch tensors.
 # Contents of this module ends up in the main namespace via _funcs.py
 # where type annotations are used in conjunction with the @normalizer decorator.
 
-
+import operator
 from typing import Optional, Sequence
 
 import torch
@@ -34,6 +34,14 @@ def copy(a: ArrayLike, order="K", subok: SubokLike = False):
     if order != "K":
         raise NotImplementedError
     return a.clone()
+
+
+def copyto(dst: NDArray, src: ArrayLike, casting="same_kind", where=NoValue):
+    if where is not NoValue:
+        raise NotImplementedError
+    (src,) = _util.typecast_tensors((src,), dst.tensor.dtype, casting=casting)
+    dst.tensor.copy_(src)
+
 
 
 def atleast_1d(*arys: ArrayLike):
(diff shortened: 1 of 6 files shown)


torch_np/_funcs_impl.py
```diff
@@ -987,6 +996,65 @@ def tile(A: ArrayLike, reps):
...
+        return torch.zeros(new_shape, dtype=a.dtype)
+
+    repeats = -(-new_size // a.numel())  # ceil division
+    a = concatenate((a,) * repeats)[:new_size]
+
+    return reshape(a, new_shape)
+
+
+def _ndarray_resize(a: ArrayLike, new_shape, refcheck=False):
+    # implementation of ndarray.resize.
```
lezcano, line 1027: my my...

lezcano (approved):

torch_np/_funcs_impl.py
lezcano: (I wanted to reference the next line)

ev-br:
Am folding it into gh-105 to bring `copyto` to address review comments.

torch_np/_funcs_impl.py
```diff
@@ -36,6 +38,13 @@ def copy(a: ArrayLike, order="K", subok: SubokLike = False):
     return a.clone()
 
 
+def copyto(dst: NDArray, src: ArrayLike, casting="same_kind", where=NoValue):
+    if where is not NoValue:
+        raise NotImplementedError
+    (src,) = _util.typecast_tensors((src,), dst.tensor.dtype, casting=casting)
+    dst.tensor.copy_(src)
```
lezcano, line 45: Actually, this should be `dst.copy_(src)`?

ev-br: Not really. `dst` is annotated as NDArray and those do not decay to torch.Tensors. src is an ArrayLike, so it can be a list to be normalized to a tensor. But `dst` cannot be a generic ArrayLike and must be an ndarray.

lezcano: Sure, it must be an ndarray, but at this point, after the normalizer, we should have just PyTorch-land objects. In particular, we should never get `tnp.ndarray`s of any kind, just Tensors.

ev-br: `dst: NDArray` annotation only checks that its argument is an ndarray exactly and does not convert it to a Tensor:

https://github.com/Quansight-Labs/numpy_pytorch_interop/blob/main/torch_np/_normalizations.py#L62

lezcano: Yeah. My point then is that `NDArray` should convert it to a `Tensor`. Why would it not do so?

ev-br: Why would it. It's a rare annotation for when need to only accept an ndarray not a general array_like. And then we know it has a `.tensor` attribute. It's a rare annotation, currently used here and for the out= arrays, which also need to be exactly arrays.

lezcano: It currently breaks the invariant that every input to these functions is something that lives in PyTorch-land. I would expect this to be a more constrained version of `ArrayLike`, but currently it has other semantics, which is a bit odd.

ev-br: Okay. So does it hold gh-105?

lezcano: It can be done afterwards as a small ENH.

ev-br: https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/108

Issue #97, opened 2023-03-31
Fix type promotion in ufuncs


ev-br:
`out=` dtype should not participate in deciding on what dtype to cast inputs to. So, `dtype=` argument forces the cast of inputs, and `out=` dtype forces the cast of `result`.

PR #107, opened 2023-04-05
BUG: fix dtype handling in ufuncs

ev-br:
Mimic numpy: dtype= argument forces the cast of inputs, and out= argument forces the cast (and broadcast) of the result.

This way, `out=` array never plays any role in dispatching. 
This is still somewhat simplified as compared to numpy since we do not have explicit loop types. So rely on pytorch doing the right thing under the hood and see if issues surface in real world testing. 

Closes gh-97.

### DIFF
--- a/torch_np/_ufuncs.py
+++ b/torch_np/_ufuncs.py
@@ -3,7 +3,28 @@ from typing import Optional
 import torch
 
 from . import _binary_ufuncs_impl, _helpers, _unary_ufuncs_impl
+from ._detail import _dtypes_impl, _util
 from ._normalizations import ArrayLike, DTypeLike, NDArray, SubokLike, normalizer
+
+
+def _ufunc_preprocess(tensors, where, casting, order, dtype, subok, signature, extobj):
+    if order != "K" or not where or signature or extobj:
+        raise NotImplementedError
+
+    if dtype is None:
+        dtype = _dtypes_impl.result_type_impl([t.dtype for t in tensors])
+
+    tensors = _util.typecast_tensors(tensors, dtype, casting)
+
+    return tensors
+
+
+def _ufunc_postprocess(result, out, casting):
+    if out is not None:
+        (result,) = _util.typecast_tensors((result,), out.dtype.torch_dtype, casting)
+        result = torch.broadcast_to(result, out.shape)
+    return result
+
 
 # ############# Binary ufuncs ######################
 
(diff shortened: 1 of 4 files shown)


torch_np/_ufuncs.py
```diff
@@ -107,12 +129,14 @@ def divmod(
...
-        (x1, x2), out, True, casting, order, dtype, subok, signature, extobj
+    tensors = _ufunc_preprocess(
+        (x1, x2), True, casting, order, dtype, subok, signature, extobj
     )
 
-    result = _binary_ufuncs_impl.divmod(*tensors)
+    quot, rem = _binary_ufuncs_impl.divmod(*tensors)
 
+    quot = _ufunc_postprocess(quot, out, casting)
+    rem = _ufunc_postprocess(rem, out, casting)
```
lezcano, line 139: These should be out1 and out2 respectively.

torch_np/tests/test_ufuncs_basic.py
```diff
@@ -105,12 +105,6 @@ def test_x_and_out_broadcast(self, ufunc):
     (np.add, operator.__add__, operator.__iadd__),
     (np.subtract, operator.__sub__, operator.__isub__),
     (np.multiply, operator.__mul__, operator.__imul__),
-    (np.divide, operator.__truediv__, operator.__itruediv__),
-    (np.floor_divide, operator.__floordiv__, operator.__ifloordiv__),
-    (np.float_power, operator.__pow__, operator.__ipow__),
-    ##   (np.remainder, operator.__mod__, operator.__imod__),   # does not handle complex
-    # remainder vs fmod?
-    # pow vs power vs float_power
```
lezcano, line 113: What happened to these?

ev-br: It's just that the smoke test parametrized on these is too simplistic to catch that `integer / integer -> float` (algebraic fields vs rings?). More comprehensive testing of these things is in `test_ufuncs` in numpy, and porting it is on a TODO list.

Since these all are now generated by the same loop, the parametrization is really not essential I think.

lezcano: fair enough

torch_np/_ufuncs.py
ev-br: Thanks, indeed.

lezcano (approved):

PR #109, opened 2023-04-06
BUG: fix nout=2 ufuncs

ev-br:
A minimal fix for a bug in handling of `out=(ndarray, ndarray)`.

### DIFF
--- a/torch_np/_ufuncs.py
+++ b/torch_np/_ufuncs.py
@@ -31,7 +31,7 @@ def _ufunc_postprocess(result, out, casting):
 _binary = [
     name
     for name in dir(_binary_ufuncs_impl)
-    if not name.startswith("_") and name not in ["torch", "matmul"]
+    if not name.startswith("_") and name not in ["torch", "matmul", "divmod"]
 ]
 
 
@@ -106,6 +106,7 @@ def matmul(
 #
 # nin=2, nout=2
 #
+@normalizer
 def divmod(
     x1: ArrayLike,
     x2: ArrayLike,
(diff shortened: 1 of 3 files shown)


torch_np/_ufuncs.py
```diff
@@ -122,15 +123,18 @@ def divmod(
...
-    if out is not None:
+    else:
         out1, out2 = out
-    if out1.shape != out2.shape or out1.dtype != out2.dtype:
-        raise ValueError("out1, out2 must be compatible")
+        if num_outs == 2:
+            if out1 is not None or out2 is not None:
+                raise TypeError(
+                    "cannot specify 'out' as both a positional and keyword argument"
+                )
```
lezcano, line 137: I don't thin this logic is correct. If `num_outs == 2` and `out = (None, None)`, here we are setting setting `out1` and `out2` to `None`.

lezcano (changes requested):

torch_np/_ufuncs.py
ev-br: Indeed, thanks for spotting this! Fixed in https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/109/commits/a855a3710ba67baf2a68342f4b1d4ef8c3398916. The relevant test needs to stay xfailed for a while, since we have hardcoded the name "out" for postprocessing. 

The current state is that `out=tuple` works, and there is an xfailed test.

lezcano (approved):
Nice. Note that before, since we were not using `normalize`, we were not casting x1 and x2 here. I guess that that happened to work because this function is a composite of other functions, but yeah...

ev-br:
Let's get this in then --- am adding `out1=, out2=` to the list of current deficiencies, gh-73

Issue #111, opened 2023-04-10
The use of `tuple` in annotations is Python 3.9


lezcano:
See https://docs.python.org/3/library/typing.html#typing.Tuple. Accordingly, we should `from __future__ import annotations`. Now, interestingly enough, when doing that in the impl files, the annotated type of the variable returned by `inspect.signature["param"].annotation` happens to be a string... 
Not sure what's going on there (perhaps @rgommers knows?), but we should test all this with Python 3.8 as it's the oldest Python that PyTorch supports.

lezcano:
I wanted to use the import as it's more future proof. `typing.Tuple` is deprecated as of 3.9.

rgommers:
that doesn't ring any bells, sorry. I'd have thought that this should "just work" with `from __future__ import annotations`

lezcano:
Changing the dict to strings and adding the `from future import...` to all the files that use the annotations seems like the right approach. I might send a PR later this week with this change and a few other proposals if I find the time.

PR #113, opened 2023-04-16
use python 3.8+

ev-br:
Add `from __future__ import annotations` to allow `tuple[ArrayLike]` typing on python 3.8, tweak the normalization internals to match PEP-563.

On CI, downgrade python version to 3.8, unpin pytorch to test on the current released version. We'll get the full pytorch CI matrix at upstreaming time, so for now I think we can invest limited effort in the CI.

fixes gh-111


lezcano (approved):
Great! Thank you for the comprehensive fix!

PR #114, opened 2023-04-16
MAINT: handle not implemented arguments centrally, via a dedicated annotation

ev-br:
as discussed in https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/112#discussion_r1162861446

This does seem to gain a marginal LOC reduction and some possible readability improvement. Two points:

- I'm {ab,re}using the `NotImplemented` singleton as the annotation. Introducing a new `NotImplementedType` is a bit awkward since `typing.NotImplementedType` is a thing from python 3.10, and we need to cover 3.8+. A simple alternative is to just use `type(NotImplemented)` instead of `NotImplemented`. Don't have a firm opinion, am open to suggestions.

- NumPy uses a weird `_NoValue` sentinel in some cases; we mimicked it by defining an alias `NoValue = None`. In our usage we typically raise NotImplementedError on things with the `NoValue` default. Now that we have a dedicated annotation, `NoValue` alias seems extraneous. Anyone sees a need for it?

### DIFF
--- a/torch_np/_ufuncs.py
+++ b/torch_np/_ufuncs.py
@@ -6,13 +6,10 @@ import torch
 
 from . import _binary_ufuncs_impl, _helpers, _unary_ufuncs_impl
 from ._detail import _dtypes_impl, _util
-from ._normalizations import ArrayLike, DTypeLike, OutArray, SubokLike, normalizer
+from ._normalizations import ArrayLike, DTypeLike, OutArray, normalizer
 
 
 def _ufunc_preprocess(tensors, where, casting, order, dtype, subok, signature, extobj):
-    if order != "K" or not where or signature or extobj:
-        raise NotImplementedError
-
     if dtype is None:
         dtype = _dtypes_impl.result_type_impl([t.dtype for t in tensors])
 
@@ -54,7 +51,7 @@ def deco_binary_ufunc(torch_func):
         casting="same_kind",
         order="K",
         dtype: DTypeLike = None,
-        subok: SubokLike = False,
+        subok: NotImplemented = False,
         signature=None,
         extobj=None,
     ):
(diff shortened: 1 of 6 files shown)


torch_np/_ufuncs.py
```diff
@@ -6,13 +6,10 @@
 
 from . import _binary_ufuncs_impl, _helpers, _unary_ufuncs_impl
 from ._detail import _dtypes_impl, _util
-from ._normalizations import ArrayLike, DTypeLike, OutArray, SubokLike, normalizer
+from ._normalizations import ArrayLike, DTypeLike, OutArray, normalizer
 
 
 def _ufunc_preprocess(tensors, where, casting, order, dtype, subok, signature, extobj):
```
lezcano, line 12: nb. It'd be better to use the `normalizer` decorator directly when defining the `@wrapped` function, to be more explicit that they're being normalized. It's happened to me a couple times that I forgot that these functions are normalized just find that they are being normalized after calling `decor_{unary,binary}`.

lezcano (commented):
I really like this! It really shows that, either you should be processing an arg in the implementation, or you should mark it as `NotImplemented`, so there's no ambiguity for "did you forget to use this arg, or is it just not implemented?"

About the `_NoValue`, it makes sense to remove it and simply replace it by `None`.

About the `NotImplementedType`, I'd say it's best to have the usual pattern of

```python
try:
  from typing import NotImplementedType
except ImportError:
  NotImplementedType = typing.TypeVar("NotImplementedType")
```
The reason is that `NotImplemented` is an object, not a class, so using it for typing is a bit odd. WDYT?

torch_np/_ufuncs.py
ev-br:
```
def decorator(...):
    @another_decorator
    def wrapped(...):
        ....
```

"I put a decorator inside your decorator", wow :-). Readability counts, so done in the updated PR.

lezcano (approved):
Great, thank you!

PR #115, opened 2023-04-17
Allow setting default dtypes

ev-br:
Add `tnp.set_default_dtype` to set the default to either follow numpy (float64) or pytorch (float32). Make it forward-facing to accept either holistic "numpy" / "pytorch" settings or explicit dtypes.

To simplify things, finish up nuking the `_details` subpackage and move its contents to the top-level torch_np folder.


torch_np/_dtypes.py
```diff
@@ -343,21 +343,40 @@ def __setstate__(self, value):
...
-    torch_dtype = dtype(dtyp).torch_dtype
-    return _dtypes_impl.get_default_type_for(torch_dtype)
+def set_default_dtype(fp_dtype="numpy", int_dtype="numpy"):
+    """Set the (global) defaults for fp and int dtypes.
+
+    Parameters
+    ----------
+    fp_dtype
+        Allowed values are "numpy", "pytorch" or dtype_like things which
+        can be converted into a tnp.DType instance.
```
lezcano, line 353: `tnp` is not a standard name. If it accepts PyTorch dtypes let's just say that. No need to be too general.

torch_np/_dtypes_impl.py
```diff
@@ -0,0 +1,74 @@
...
+#### defaults : mimic NumPy, allow user control
+
+
+class DefaultDTypes:
+    """A bag class for dtype defaults."""
+
+    def __init__(self, fp_dtype="numpy", int_dtype="pytorch"):
+
+        if not (fp_dtype in ["numpy", "pytorch"] or isinstance(fp_dtype, torch.dtype)):
+            raise TypeError(f"failed to interpter {fp_dtype} as torch.dtype.")
```
lezcano, line 17: nit interpret. Same below.

torch_np/_dtypes_impl.py
```diff
@@ -0,0 +1,74 @@
...
+"""
+import torch
+
+#### defaults : mimic NumPy, allow user control
+
+
+class DefaultDTypes:
+    """A bag class for dtype defaults."""
+
+    def __init__(self, fp_dtype="numpy", int_dtype="pytorch"):
```
lezcano, line 14: Even if it's equivalent, let's set both to `"numpy"`, as in `set_default_dtype`.

lezcano (commented):
At the moment, we have the processing of the dtypes args partly in the `set_default_dtype` func, and partly in the `DefaultDTypes`. I think it would be better to have it all in `set_default_dtype`. That way it's easier to locate and understand. Then, if this is done this way, `DefaultDTypes` can simply be a `namedtuple` of `torch` dtypes.

torch_np/_dtypes.py
ev-br: `tnp` is gone in the update, but it's not only pytorch dtypes: it also needs to accept string representations 'float32',  and other numpy-isms (dtype="f8") etc. These conversions are taken care of in `dtype(fp_dtype)` calls.

ev-br:
Updated to take into account review comments.

lezcano (approved):
Great, thanks!

torch_np/_dtypes.py
```diff
@@ -343,21 +343,65 @@ def __setstate__(self, value):
...
+    The old default dtype state: a namedtuple with attributes ``float_dtype``,
+    ``complex_dtypes`` and ``int_dtype``. These attributes store *pytorch*
+    dtypes.
+
+    Notes
+    ------------
+    This functions has a side effect: it sets the global state with the provided dtypes.
+
+    The complex dtype has bit width of at least twice the width of the float
+    dtype, i.e. it's complex128 for float64 and complex64 for float64.
```
lezcano, line 371:
```suggestion
    dtype, i.e. it's complex128 for float64 and complex64 for float32.
```

torch_np/_dtypes.py
```diff
@@ -343,21 +343,40 @@ def __setstate__(self, value):
...
-
-def default_complex_type():
-    return dtype(_dtypes_impl.default_complex_dtype)
-
-
-def get_default_dtype_for(dtyp):
-    torch_dtype = dtype(dtyp).torch_dtype
-    return _dtypes_impl.get_default_type_for(torch_dtype)
+def set_default_dtype(fp_dtype="numpy", int_dtype="numpy"):
+    """Set the (global) defaults for fp and int dtypes.
```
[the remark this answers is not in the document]
lezcano: Actually, it'd be good to be more explicit at the top of the docs, rather than having people looking for the actual information.
```suggestion
    """Set the (global) defaults for fp, complex, and int dtypes.
    
    The complex dtype is inferred from the float (fp) dtype. It has 
    a width at least twice the width of the float dtype, 
    i.e., it's complex128 for float64 and complex64 for float32.
```

torch_np/_dtypes.py
```diff
@@ -343,21 +343,69 @@ def __setstate__(self, value):
...
+    The old default dtype state: a namedtuple with attributes ``float_dtype``,
+    ``complex_dtypes`` and ``int_dtype``. These attributes store *pytorch*
+    dtypes.
+
+    Notes
+    ------------
+    This functions has a side effect: it sets the global state with the provided dtypes.
+
+    The complex dtype has bit width of at least twice the width of the float
+    dtype, i.e. it's complex128 for float64 and complex64 for float32.
```
lezcano, line 375: In the end this was repeated in two places. We can probably delete this bit in any other PR

PR #116, opened 2023-04-17
`tnp.put()` + testing

honno:
`put()` implementation and my own test. This was tricky but I think it's pretty robust, sans the TODO

* [x] support non-array/tensor indices in `ind`

Notably in `test_put` I'm testing `tnp.put()` output against NumPy's own `np.put()`, which as you can see is a bit noisy right now due to `tnp.asarray()` not carrying over dtypes (yet? something I'd like to fix as it enables future "comparison testing").

### DIFF
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -904,8 +904,10 @@ def put(
     # If ind is larger than v, broadcast v to the would-be resulting shape. Any
     # unnecessary trailing elements are then trimmed.
     if ind.numel() > v.numel():
-        result_shape = torch.broadcast_shapes(v.shape, ind.shape)
-        v = torch.broadcast_to(v, result_shape)
+        ratio = (ind.numel() + v.numel() - 1) // v.numel()
+        sizes = [ratio]
+        sizes.extend([1 for _ in range(v.dim() - 1)])
+        v = v.repeat(*sizes)
     # Trim unnecessary elements, regarldess if v was broadcasted or not. Note
     # np.put() trims v to match ind by default too.
     if ind.numel() < v.numel():
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -901,14 +901,12 @@ def put(
     mode: NotImplementedType = "raise",
 ):
     v = v.type(a.dtype)
-    # If ind is larger than v, broadcast v to the would-be resulting shape. Any
+    # If ind is larger than v, expand v to at least the size of ind. Any
     # unnecessary trailing elements are then trimmed.
     if ind.numel() > v.numel():
         ratio = (ind.numel() + v.numel() - 1) // v.numel()
-        sizes = [ratio]
-        sizes.extend([1 for _ in range(v.dim() - 1)])
-        v = v.repeat(*sizes)
-    # Trim unnecessary elements, regarldess if v was broadcasted or not. Note
+        v = v.unsqueeze(0).expand((ratio,) + v.shape)
+    # Trim unnecessary elements, regarldess if v was expanded or not. Note
     # np.put() trims v to match ind by default too.
     if ind.numel() < v.numel():
         v = v.flatten()
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -895,7 +895,7 @@ def take_along_axis(arr: ArrayLike, indices: ArrayLike, axis):
 
 
 def put(
-    a: ArrayLike,
+    a: NDArray,
     ind: ArrayLike,
     v: ArrayLike,
     mode: NotImplementedType = "raise",
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -896,29 +896,15 @@ def take_along_axis(arr: ArrayLike, indices: ArrayLike, axis):
 
 def put(
     a: ArrayLike,
-    ind: Sequence[ArrayLike],
+    ind: ArrayLike,
     v: ArrayLike,
     mode: NotImplementedType = "raise",
 ):
-    indexes = list(ind)
-    for i, index in enumerate(indexes):
-        if not isinstance(index, torch.Tensor):
-            indexes[i] = torch.as_tensor(index)
-    index = torch.concat(indexes)
-    index[index < 0] += a.numel()  # normalise negative indices
-    index_u, index_c = torch.unique(index, return_counts=True)
-    duplicated_indices = index_u[index_c > 1]
-    if duplicated_indices.numel() > 0:
-        raise NotImplementedError(
-            "duplicated indices are not supported. duplicated indices: "
-            f"{duplicated_indices}"
-        )
-    source = v
-    if source.numel() < index.numel():
-        numel_ratio = float(index.numel() / source.numel())
+    if v.numel() < ind.numel():
+        numel_ratio = float(ind.numel() / v.numel())
         if numel_ratio.is_integer():
-            source = torch.stack([source for _ in range(int(numel_ratio))])
-    a.put_(index, source)
+            v = torch.stack([v for _ in range(int(numel_ratio))])
+    a.put_(ind, v)
     return None
 
 
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -900,10 +900,11 @@ def put(
     v: ArrayLike,
     mode: NotImplementedType = "raise",
 ):
-    if v.numel() < ind.numel():
-        numel_ratio = float(ind.numel() / v.numel())
-        if numel_ratio.is_integer():
-            v = torch.stack([v for _ in range(int(numel_ratio))])
+    numel_ratio = ind.numel() / v.numel()
+    if numel_ratio.is_integer():
+        sizes = [int(numel_ratio)]
+        sizes.extend([1 for _ in range(v.dim() - 1)])
+        v = v.repeat(*sizes)
     a.put_(ind, v)
     return None
 
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -900,6 +900,7 @@ def put(
     v: ArrayLike,
     mode: NotImplementedType = "raise",
 ):
+    v = v.type(a.dtype)
     numel_ratio = ind.numel() / v.numel()
     if numel_ratio.is_integer():
         sizes = [int(numel_ratio)]
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -894,6 +894,28 @@ def take_along_axis(arr: ArrayLike, indices: ArrayLike, axis):
     return torch.take_along_dim(arr, indices, axis)
 
 
+def put(
+    a: ArrayLike,
+    ind: Sequence[ArrayLike],
+    v: ArrayLike,
+    mode: NotImplementedType = "raise",
+):
+    index = torch.concat(ind)
+    index[index < 0] += a.numel()  # normalise negative indices
+    index_u, index_c = torch.unique(index, return_counts=True)
+    duplicated_indices = index_u[index_c > 1]
+    if duplicated_indices.numel() > 0:
+        raise NotImplementedError(
+            "duplicated indices are not supported. duplicated indices: "
+            f"{duplicated_indices}"
+        )
+    source = v
+    if source.numel() < index.numel():
+        source = torch.broadcast_to(source, index.size())
+    a.put_(index, source)
+    return None
+
+
 def put_along_axis(arr: ArrayLike, indices: ArrayLike, values: ArrayLike, axis):
     (arr,), axis = _util.axis_none_ravel(arr, axis=axis)
     axis = _util.normalize_axis_index(axis, arr.ndim)
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -900,7 +900,11 @@ def put(
     v: ArrayLike,
     mode: NotImplementedType = "raise",
 ):
-    index = torch.concat(ind)
+    indexes = list(ind)
+    for i, index in enumerate(indexes):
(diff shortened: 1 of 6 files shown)


torch_np/_funcs_impl.py
```diff
@@ -913,6 +913,28 @@ def take_along_axis(arr: ArrayLike, indices: ArrayLike, axis):
...
+    if duplicated_indices.numel() > 0:
+        raise NotImplementedError(
+            "duplicated indices are not supported. duplicated indices: "
+            f"{list(duplicated_indices)}"
+        )
+    source = v
+    if source.numel() < index.numel():
+        source = torch.broadcast_to(source, index.size())
+    # Note Tensor.put_ acts in-place, while Tensor.put (no trailing underscore)
+    # is seemingly out-of-place.
```
lezcano, line 933: This is a convention all throughout PyTorch.

torch_np/_funcs_impl.py
```diff
@@ -913,6 +913,28 @@ def take_along_axis(arr: ArrayLike, indices: ArrayLike, axis):
     return torch.take_along_dim(arr, indices, axis)
 
 
+def put(a: ArrayLike, ind: Sequence[ArrayLike], v: ArrayLike, mode="raise"):
+    if mode != "raise":
+        raise NotImplementedError(f"{mode=}")
+
+    index = torch.concat(ind)
```
lezcano, line 920: According to NumPy's docs this is an array-like, so why using a sequence?

torch_np/_funcs_impl.py
```diff
@@ -913,6 +913,28 @@ def take_along_axis(arr: ArrayLike, indices: ArrayLike, axis):
...
+    index[index < 0] += a.numel()  # normalise negative indices
```
lezcano, line 921: PyTorch supports negative indices, so this should not be necessary.

torch_np/_funcs_impl.py
```diff
@@ -913,6 +913,28 @@ def take_along_axis(arr: ArrayLike, indices: ArrayLike, axis):
...
+    index_u, index_c = torch.unique(index, return_counts=True)
+    duplicated_indices = index_u[index_c > 1]
+    if duplicated_indices.numel() > 0:
+        raise NotImplementedError(
+            "duplicated indices are not supported. duplicated indices: "
+            f"{list(duplicated_indices)}"
+        )
```
lezcano, line 928: Data-dependent flow is not great for compilers, so we try to avoid it when it's just used to throw an error.

torch_np/_funcs_impl.py
```diff
@@ -913,6 +913,28 @@ def take_along_axis(arr: ArrayLike, indices: ArrayLike, axis):
     return torch.take_along_dim(arr, indices, axis)
 
 
+def put(a: ArrayLike, ind: Sequence[ArrayLike], v: ArrayLike, mode="raise"):
+    if mode != "raise":
+        raise NotImplementedError(f"{mode=}")
```
lezcano, line 918: This can probably be implemented via https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/114

lezcano (commented):
This looks good, but I believe it may be possible to simplify some bits of the implementation.

torch_np/_funcs_impl.py
honno: Ah good shout, had thought of a few edge cases but didn't realise how powerful the `ArrayLike` normalisation is! So now using `ArrayLike` which clears up `put()` nicely and keeps `test_put` green :tada:

torch_np/_funcs_impl.py
honno: (This was for making unique work on equivalent negative indices, but that's gone now so resolving this)

ev-br:
Could you also try un-xfailing `test_put` and `test_put_noncontiguous` in https://github.com/Quansight-Labs/numpy_pytorch_interop/blob/main/torch_np/tests/numpy_tests/core/test_multiarray.py

torch_np/_funcs_impl.py
```diff
@@ -909,6 +909,20 @@ def take_along_axis(arr: ArrayLike, indices: ArrayLike, axis):
...
+def put(
+    a: ArrayLike,
+    ind: ArrayLike,
+    v: ArrayLike,
+    mode: NotImplementedType = "raise",
+):
+    if v.numel() < ind.numel():
+        numel_ratio = float(ind.numel() / v.numel())
+        if numel_ratio.is_integer():
+            v = torch.stack([v for _ in range(int(numel_ratio))])
```
lezcano, line 921: This logic looks a bit odd. 

So, you do this stacking (which should probably be done via `torch.repeat`) when the number of elements of v divides that of ind, and otherwise we do nothing (and I guess we error out inside `a.put_`?

Do you have a reference for where this is implemented in NumPy?

ev-br: ISTM it's this one: https://github.com/numpy/numpy/blob/main/numpy/core/src/multiarray/item_selection.c#L350

lezcano: So it seems like it also wraps when they are not divisible?
https://github.com/numpy/numpy/blob/6073588dd73809a60819d71b9527194195f73f08/numpy/core/src/multiarray/item_selection.c#L501
See also
```python
>>> import numpy as np
>>> x = np.array([1,2,3,4])
>>> x.put([-1, -2, -3], [-8, -7])
>>> x
array([ 1, -8, -7, -8])
```

lezcano: I don't think we have a function that does "repeat until you fit this shape", but you can use `torch.repeat` and then take as many elements from it as you need via advanced indexing or `index_select`. This second part can be guarded by a divisibility test, so that we just do it when it's strictly necessary.

honno: For some reason thought `int / int` could return both ints and float-ints, hence the unnecessary `float()` casting. Removed!

> you can use `torch.repeat` and then take as many elements from it as you need via advanced indexing or `index_select`

I don't quite get this sorry—as I read it, you'd need to workout a repeat which is the same-size/bigger-than the indexer anywho. Pushed a change, lemme know what you think.

<br />
<sup>
<blockquote>Do you have a reference for where this is implemented in NumPy?</blockquote>

Haven't been checking the NumPy implementations for these things, should start I suppose :sweat_smile:
</sup>

torch_np/tests/numpy_tests/core/test_multiarray.py
```diff
@@ -2667,14 +2667,10 @@ def test_put(self):
...
 
-        # check must be writeable
-        a = np.zeros(6)
-        a.flags.writeable = False
-        assert_raises(ValueError, a.put, [1, 3, 5], [1, 3, 5])
-
         # when calling np.put, make sure a
         # TypeError is raised if the object
         # isn't an ndarray
+        pytest.xfail("XXX: Argument normalisation prevents catching this")
```
ev-br: Does it mean that it needs to be annotated as NDArray, not ArrayLike?

honno: Ah I see yeah, thanks.

torch_np/_funcs_impl.py
```diff
@@ -909,6 +909,22 @@ def take_along_axis(arr: ArrayLike, indices: ArrayLike, axis):
...
+    ind: ArrayLike,
+    v: ArrayLike,
+    mode: NotImplementedType = "raise",
+):
+    v = v.type(a.dtype)
+    numel_ratio = ind.numel() / v.numel()
+    if numel_ratio.is_integer():
+        sizes = [int(numel_ratio)]
+        sizes.extend([1 for _ in range(v.dim() - 1)])
+        v = v.repeat(*sizes)
```
lezcano, line 923: A minor: When you want to see if an int divides another you do `a % b == 0`.
Now, this code still does nothing when `ind` and `v` do not have the same number of elements. I think the correct implementation would go along the lines of:
```suggestion
    # Nb. If ind.numel > v.numel() and v.numel() doesn't divide ind.numel()
    #     we're going to enter the first if, create more elements than necessary,
    #     and then trim the unnecessary elements in the second if
    if ind.numel() > v.numel():
        # Repeat v
        # ceil(ind.numel() / v.numel())
        ratio = (ind.numel() + v.numel() - 1) // v.numel();
        v = v.repeat(v.shape + (ratio,))
    if ind.numel() < v.numel():
        # Trim the last elements
        if v.ndim > 1:
            v = v.flatten()
        v = v[:ind.numel()]
```
Note that this has not been tested. Also, I doubt my code works when dealing with 0D arrays.

honno: Ah I see, makes sense! Pushed a change to this effect, notably

* I broadcasted v as opposed to repeat, as there were edge cases where repeating diverged from NumPy behaviour where broadcasting did not.
* I didn't check `if v.ndim > 1` <sup>(I imagine we would of wanted `v.ndim != 1` instead)</sup> before flattening, as I assume its a fast no-op anyway for 1D arrays?

torch_np/_funcs_impl.py
```diff
@@ -909,6 +909,27 @@ def take_along_axis(arr: ArrayLike, indices: ArrayLike, axis):
...
+    ind: ArrayLike,
+    v: ArrayLike,
+    mode: NotImplementedType = "raise",
+):
+    v = v.type(a.dtype)
+    # If ind is larger than v, broadcast v to the would-be resulting shape. Any
+    # unnecessary trailing elements are then trimmed.
+    if ind.numel() > v.numel():
+        result_shape = torch.broadcast_shapes(v.shape, ind.shape)
+        v = torch.broadcast_to(v, result_shape)
```
lezcano, line 923: Is this correct? If `ind = torch.arange(4)` and `v = torch.arange(3).unsqueeze(1)`, then we'll broadcast `v`, and we'll then trim it in the next line. The issue here is that by doing the broadcasting in the last dimension, we'll be copying the elements `0, 0, 0, 0`, rather than `0, 1, 2, 0`. For the elements to repeat as we want when doing the `flatten` below we want to do the broadcasting on the left-most dimension. This was also a bug in my proposed implementation. You can do that by using the ceil computation I proposed + `unsqueeze(0)` + `expand`.

honno: Ah yeah you're right, for reference

```python
>>> a = np.arange(5)
>>> ind = np.arange(4)
>>> v = np.arange(3).unsqueeze(1)
>>> np.put(a, ind, v)
>>> a
array([0, 1, 2, 0, 4])
```

```python
>>> x = tnp.arange(5)
>>> ind = tnp.arange(4)
>>> v = tnp.asarray([[0], [1], [2]])
>>> tnp.put(x, ind, v)
>>> x
array_w([0, 0, 0, 0, 4])
```

So Ill need to do more exploration/testing which still satisfies the existing problem I came across. Will leave for next week :sweat_smile:

<details>
<summary>FWIW dirty/un-minified failing case I alluded to before when using repeat scenario before</summary>

```python
     AssertionError: 
     Arrays are not equal
     
     Mismatched elements: 1 / 2 (50%)
      x: array_w([[[ True, False]]])
      y: array_w([[[True, True]]])
     Falsifying example: test_put(
         np_x=array([[[False, False]]]),
         data=data(...),
     )
     Draw 1 (ind): (array([[-1],[-1]]), array([[0],[0]]), array([[0],[0]]))
     Draw 2 (v): array([[False],
            [ True]])
     (after put) np_x=array([[[ True,  True]]])
     tnp_x=array_w([[[False, False]]])
     (after put) tnp_x=array_w([[[ True, False]]])


     Mismatched elements: 2 / 12 (16.7%)
      x: array_w([[ True, False, False,  True],
             [False, False, False, False],
             [False, False, False, False]])
      y: array_w([[ True, False,  True, False],
             [False, False, False, False],
             [False, False, False, False]])
     Falsifying example: test_put(
         np_x=array([[False, False, False, False],
                [False, False, False, False],
                [False, False, False, False]]),
         data=data(...),
     )
     Draw 1 (ind): (array([[1],
             [2]]), array([[3],
             [0]]))
     Draw 2 (v): array([[False],
            [ True]])
     (after put) np_x=array([[ True, False,  True, False],
            [False, False, False, False],
            [False, False, False, False]])
     tnp_x=array_w([[False, False, False, False],
             [False, False, False, False],
             [False, False, False, False]])
     (after put) tnp_x=array_w([[ True, False, False,  True],
             [False, False, False, False],
             [False, False, False, False]])
```

</details>

torch_np/_funcs_impl.py
```diff
@@ -894,6 +894,29 @@ def take_along_axis(arr: ArrayLike, indices: ArrayLike, axis):
...
+    mode: NotImplementedType = "raise",
+):
+    v = v.type(a.dtype)
+    # If ind is larger than v, broadcast v to the would-be resulting shape. Any
+    # unnecessary trailing elements are then trimmed.
+    if ind.numel() > v.numel():
+        ratio = (ind.numel() + v.numel() - 1) // v.numel()
+        sizes = [ratio]
+        sizes.extend([1 for _ in range(v.dim() - 1)])
+        v = v.repeat(*sizes)
```
lezcano, line 910: Nit. `[1] * (v.dim() - 1)`.

That being said, as discussed inhttps://github.com/Quansight-Labs/numpy_pytorch_interop/pull/116#discussion_r1172316059, what you probably want is an unsqueeze(0) + expand, as you don't want to copy the data if you can avoid it.

torch_np/_funcs_impl.py
honno: Pushed a fix using `repeat` again, and make sure its tested for in `test_put`.

Notably repeating just on the 0th axis seems the right call for our `put()` to align with NumPy-proper, as other ways of constructing `*sizes` falls short (i.e. Hypothesis will come up with failing examples in `test_put`).

torch_np/_funcs_impl.py
honno: Ah thanks I see, pushed a change to this effect.

lezcano (approved):
Now we're cooking with gas! Thank you for the PR!

PR #117, opened 2023-04-18
Some tests on toy examples

ev-br:
Try testing our torch_np wrapper on real-world examples. Several toy problems first:

-  A toy NN from scratch using numpy, https://www.geeksforgeeks.org/implementation-of-neural-network-from-scratch-using-numpy/

Several examples from N. Rougier,  From python to numpy, https://www.labri.fr/perso/nrougier/from-python-to-numpy/

- Build a maze and find a path; : 
https://github.com/rougier/from-python-to-numpy/blob/master/code/maze_numpy.py

- Simulate a diffusion/advection spreading of a smoke ring: https://github.com/rougier/from-python-to-numpy/blob/master/code/smoke_1.py

- Construct / draw the Mandelbrot fractal: https://github.com/rougier/from-python-to-numpy/blob/master/code/mandelbrot_numpy_1.py

-------------------

The strategy is to replace `import numpy as np` with `import torch_np as np` (i.e., in eager mode). See e2e/tests.md for some notes.


lezcano (approved):
This is neat! The only comment I have (not for this PR) is that perhaps we can wire `np.random.seed` to `torch.set_seed` together with a `torch.warn_once` saying "careful, the random numbers are not the same!". This would allow us to not use `_np` at all, and still run these models. The point here is that here they just set a seed for reproducibility, but any seed (and hopefully any generator) will work just fine.

ev-br:
`_np.random` is me, not _them_. It's only to check that the results are exactly the same.

ev-br:
@honno could you give me a hand at excluding everything under `./e2e` from pre-commit please?

ev-br:
OK, a first fix to `ndarray.__setitem__`: https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/117/commits/5c8f5684b1d6ec53e3b8d9ac6c9bfdd288b23827

Here's a minimal reproducer:

```
In [48]: ts = torch.empty(3, dtype=torch.float32)

In [49]: idx = [2, 0, 1]

In [50]: tf = torch.as_tensor([2, 2, 2], dtype=float)

In [51]: ts[idx] = tf
Input In [51], in <cell line: 1>()
----> 1 ts[idx] = tf

RuntimeError: Index put requires the source and destination dtypes match, got Float for the destination and Double for the source.
```

Note that in PyTorch this is specific to advanced indexing:

```
In [52]: ts[...] = tf

In [53]: ts
Out[53]: tensor([2., 2., 2.], dtype=torch.float32)
```

So maybe this can be considered a bug in PyTorch @lezcano ?

lezcano:
Not a bug, as it throws a clean error, so this was accounted for. The fix LGTM.

ev-br:
Ok, we have four examples, which all basically work (in eager mode). Let's merge it, to have a link to add to the RFC.

lezcano:
This is great! Could you add a short readme, similar to the OP, in the folder?

ev-br:
Is this what you meant: https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/120

ev-br:
Nothing really interesting beyond "sloth slow in eager mode". How slow exactly I did not time, would be more interesting when integrated with torch.dynamo, I guess.

And it's not mine really, all credit goes to Nicholas Rougier :-).

PR #119, opened 2023-04-19
Fix fill_diagonal

lezcano:
fill_diagonal had an implementation that assumed that `.ravel()` returns a view onto the original tensor. This is not true, as there are many tensors that cannot be flattened. `x.flat` in NumPy returns an iterator that iterates the elements "in order". We do not have anything like that in PyTorch. This PR implements this operation via some rather funky advanced indexing. The `wrap` kwarg is... weird to say the least...


lezcano:
Note that this sort of arange-based indexing is compiled into efficient code by inductor, which uses the loop variables (or the thread_idx) to simulate the elements of `arange`.

PR #121, opened 2023-04-20
Assorted fixes and simplifications

lezcano:
- Give up trying to imitate the NumPy exceptions altogether
- Remove duplicated round, real imag
- Remove the implementation of nanmean, as it had a number of issues. If we implement masked reducitons, we should probably do them all in one go, reusing the aux functions we used for reductions.
- Minor cosmetic changes

### DIFF
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -71,14 +71,16 @@ def atleast_3d(*arys: ArrayLike):
 
 
 def _concat_check(tup, dtype, out):
+    if tup == ():
+        raise ValueError("need at least one array to concatenate")
+
     """Check inputs in concatenate et al."""
-    if out is not None:
-        if dtype is not None:
-            # mimic numpy
-            raise TypeError(
-                "concatenate() only takes `out` or `dtype` as an "
-                "argument, but both were provided."
-            )
+    if out is not None and dtype is not None:
+        # mimic numpy
+        raise TypeError(
+            "concatenate() only takes `out` or `dtype` as an "
+            "argument, but both were provided."
+        )
 
 
 def _concat_cast_helper(tensors, out=None, dtype=None, casting="same_kind"):
@@ -343,6 +345,11 @@ def arange(
 
     # work around RuntimeError: "arange_cpu" not implemented for 'ComplexFloat'
     work_dtype = torch.float64 if target_dtype.is_complex else target_dtype
+
+    if (step > 0 and start > stop) or (step < 0 and start < stop):
+        # empty range
+        return torch.empty(0, dtype=target_dtype)
+
     result = torch.arange(start, stop, step, dtype=work_dtype)
     result = _util.cast_if_needed(result, target_dtype)
     return result
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -14,7 +14,7 @@ from typing import Optional, Sequence
 
 import torch
 
-from . import _dtypes_impl, _helpers
+from . import _dtypes_impl
 from . import _reductions as _impl
 from . import _util
 from ._normalizations import (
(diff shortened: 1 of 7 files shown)


torch_np/_funcs_impl.py
```diff
@@ -730,50 +709,29 @@ def triu_indices(n, k=0, m=None):
 def tril_indices_from(arr: ArrayLike, k=0):
     if arr.ndim != 2:
         raise ValueError("input array must be 2-d")
-    result = torch.tril_indices(arr.shape[0], arr.shape[1], offset=k)
-    return tuple(result)
+    # Return a tensor rather than a tuple to avoid a graphbreak
+    return torch.tril_indices(arr.shape[0], arr.shape[1], offset=k)
```
ev-br, line 713: This is a break w.r.t. numpy, so should be added to the list in https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/73 I guess.
Also this should fail on CI, so those tests need to be adjusted. 

Semantics wise, should we not be worried about

```
In [40]: a = np.arange(6).reshape(3, 2)

In [41]: idx = np.tril_indices_from(a)

In [42]: a[idx]
Out[42]: array([0, 2, 3, 4, 5])

In [43]: a[np.asarray(idx)]
Out[43]: 
array([[[0, 1],
        [2, 3],
        [2, 3],
        [4, 5],
        [4, 5]],

       [[0, 1],
        [0, 1],
        [2, 3],
        [0, 1],
        [2, 3]]])
```

torch_np/_funcs_impl.py
```diff
@@ -730,50 +709,29 @@ def triu_indices(n, k=0, m=None):
...
-        result = a.nanmean(dtype=dtype)
-        if keepdims:
-            result = torch.full(a.shape, result, dtype=result.dtype)
-    else:
-        result = a.nanmean(dtype=dtype, dim=axis, keepdim=bool(keepdims))
-    if out is not None:
-        out.copy_(result)
-    return result
+def nanmean():
+    raise NotImplementedError
```
ev-br, line 734: Yeah, these all should go. This weirdness was to be able to port nanXXX tests, so should probbaly be redone in one go. This PR or a separate one, whichever works.

torch_np/_funcs_impl.py
```diff
@@ -1750,8 +1666,7 @@ def round_(a: ArrayLike, decimals=0, out: Optional[OutArray] = None):
     return result
 
 
-around = round_
-round = round_
+around = round
```
ev-br, line 1669: so we're OK with not having a `round_` alias? IUIC it might get deprecated in NumPy 2.0, so probably OK.

ev-br (commented):
Looks good modulo `tri{l,u}_indices_from` semantic break. That one might be OK, too, just noting it to make sure we want to do it. ISTM the main usage of these is to actually index something with the result, so if we return a tensor, the next stage for a user is to swear and throw an explicit `tuple()` call. So we're back to the same graph break and only earn a swear from a user?

torch_np/_funcs_impl.py
lezcano: Oh, `round_` is also a thing?! Let me add that. Leaving `round` as the main one tho.

torch_np/_funcs_impl.py
lezcano: Added to that issue.

lezcano: >ISTM the main usage of these is to actually index something with the result, so if we return a tensor, the next stage for a user is to swear and throw an explicit tuple() call. So we're back to the same graph break and only earn a swear from a user

As discussed in this example, we should be fine, as it's possible to index also with a tensor while preserving the semantics.

lezcano:
Addressed the reviews.

ev-br:
LGTM

PR #123, opened 2023-04-22
add wrappers for np.fft

ev-br:
Is mostly straightforward, as pytorch API is nearly identical indeed. The main difference seems to be that  `numpy.fft` returns 64-bit transforms for all input dtypes. We now have `set_default_dtype` though, so we probably want fft routines to react to it: 

```
>>> a = np.arange(5, dtype=np.float32)
>>> np.fft.fft(a).dtype == np.complex128            # this is the (numpy) way
True
>>> np.set_default_dtype("pytorch")
>>> np.fft.fft(a).dtype == np.complex64
```

### DIFF
--- a/torch_np/tests/numpy_tests/fft/test_pocketfft.py
+++ b/torch_np/tests/numpy_tests/fft/test_pocketfft.py
@@ -19,15 +19,16 @@ def fft1(x):
     return np.sum(x*np.exp(phase), axis=1)
 
 
-@pytest.mark.xfail(reason='TODO')
 class TestFFTShift:
 
     def test_fft_n(self):
-        assert_raises(ValueError, np.fft.fft, [1, 2, 3], 0)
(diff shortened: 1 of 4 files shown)


torch_np/tests/numpy_tests/fft/test_pocketfft.py
```diff
@@ -40,13 +41,15 @@ def test_identity(self):
                             xr[0:i], atol=1e-12)
 
     def test_fft(self):
+
+        np.random.seed(1234)
         x = random(30) + 1j*random(30)
-        assert_allclose(fft1(x), np.fft.fft(x), atol=1e-6)
-        assert_allclose(fft1(x), np.fft.fft(x, norm="backward"), atol=1e-6)
+        assert_allclose(fft1(x), np.fft.fft(x), atol=2e-5)
```
ev-br, line 47: Had to bump the tolerance. Not much we can do I guess, this fall straight through to pytorch.

lezcano (approved):
Cool!

PR #127, opened 2023-04-25
ENH: add einsum

ev-br:
A couple of notes:

- the generic normalization machinery does not handle signatures with positional args before varargs. This is `einsum` and `gradient`, so keep the generic machinery simple and do things manually here.
- einsum signature is a bit weird even in numpy, with `**kwargs`, so follow that. Not sure why it is what it is, one possible reason could be that keyword-only args after varargs is a syntax error (?) E.g. `def func(a, *args, *, out=None): pass`.  
- IIUC, pytorch does not allow the same level of control over the `optimize = {False, True, ‘greedy’, ‘optimal’}` argument, so silenty ignore this argument

Also while at it, add a dedicated annotation to validate the `casting=...` argument. This is strictly speaking extraneous to this PR, so can take it out if desired.


torch_np/_funcs_impl.py
```diff
@@ -1208,6 +1228,68 @@ def outer(a: ArrayLike, b: ArrayLike, out: Optional[OutArray] = None):
...
+
+    # parse arrays and normalize them
+    if isinstance(operands[0], str):
+        # ("ij->", arrays) format
+        sublist_format = False
+        subscripts, array_operands = operands[0], operands[1:]
+    else:
+        # op, str, op, str ... format: normalize every other argument
+        sublist_format = True
+        array_operands = operands[:-1][::2]
```
lezcano, line 1269: Why the `[:-1]`? Isn't `sublistout` optional? Also, don't we want to preprocess that one as well, perhaps asserting that it's an `ndarray`?

torch_np/_funcs_impl.py
```diff
@@ -1208,6 +1228,68 @@ def outer(a: ArrayLike, b: ArrayLike, out: Optional[OutArray] = None):
...
+        if dtype is None
+        else dtype
+    )
+    tensors = _util.typecast_tensors(tensors, target_dtype, casting)
+
+    if sublist_format:
+        # recombine operands
+        sublists = operands[1::2]
+        sublistout = (operands[-1],) if len(operands) % 2 == 1 else ()
+        operands = builtins.sum((_ for _ in zip(tensors, sublists)), ()) + sublistout
```
lezcano, line 1283: It took me a bit of thinking to understand what was going on here. What about the more explicit using `chain`
```suggestion
        operands = list(itertools.chain(*zip(tensor, operands[1::2])))
        if len(operands) % 2 == 1:
            operands.append(operands[-1])
```

lezcano (commented):
I can just imagine the pain that was implementing this one. I left a few comments, but overall looks good.

torch_np/_funcs_impl.py
ev-br: Have to admit I kind of liked the _what the ..._ vibe of the `sum(..., ())` trick :-). Anyway, changed with some small tweaks for readability.

torch_np/_funcs_impl.py
ev-br: Exactly! 
- If sublistout is not given, the length of `operands` is even, and we pick odd-numbered elements, which are arrays. 
- If sublistout is given, the length of `operands` is odd, we peel off the last one, and pick odd-numbered elements, which are arrays. Without `[:-1]`, we would have picked `sublistout`, too --- and it's a sublist, not an array.

And, no, it's not an array really.  Can contain e.g. an Ellipsis:

```
            assert_equal(np.einsum("...i->...", a, optimize=do_opt),
                         np.sum(a, axis=-1).astype(dtype))
            assert_equal(np.einsum(a, [Ellipsis, 0], [Ellipsis], optimize=do_opt),
                         np.sum(a, axis=-1).astype(dtype))
```

torch_np/_funcs_impl.py
```diff
@@ -1208,6 +1228,68 @@ def outer(a: ArrayLike, b: ArrayLike, out: Optional[OutArray] = None):
...
+
+    parm = lambda _: None  # a fake duck-typed inspect.Parameter stub
+    parm.name = "out"
+    out = normalize_outarray(out, parm=parm)
+
+    parm.default = "K"
+    parm.name = "order"
+    order = normalize_not_implemented(kwargs.pop("order", "K"), parm=parm)
+    if kwargs:
+        raise TypeError("unknown arguments: ", kwargs)
```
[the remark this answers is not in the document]
ev-br: Oh. Now that I re-read https://pytorch.org/docs/stable/generated/torch.einsum.html, there is `torch.backends.opt_einsum = "greedy"` so we can support this :-). Question: is there a way of controlling `torch.backends` in a context manager, other than a `try... finally` block?

torch_np/_funcs_impl.py
```diff
@@ -1208,6 +1228,68 @@ def outer(a: ArrayLike, b: ArrayLike, out: Optional[OutArray] = None):
...
+    #       casting='safe', optimize=False)
+
+    from ._normalizations import (
+        maybe_copy_to,
+        normalize_casting,
+        normalize_dtype,
+        normalize_not_implemented,
+        normalize_outarray,
+        wrap_tensors,
+    )
```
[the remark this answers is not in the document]
ev-br: If removing imports is the goal, that won't help really. These normalizers are otherwise only used in `_normalizations.py`, so these need to be imported either way.

torch_np/_funcs_impl.py
lezcano: Right. Could you please leave a comment or a link to this explanation?

torch_np/_funcs_impl.py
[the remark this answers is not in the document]
lezcano: What I'd like to remove is local imports. They are a massive red flag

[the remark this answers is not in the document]
ev-br: That would create an import cycle since `_funcs.py` imports `_funcs_impl`.

We can of course, import these at the `_funcs_impl` module level and special-case them to not get dumped into the global namespace, if that's really what you think is needed?

torch_np/_funcs_impl.py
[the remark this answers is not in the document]
ev-br: Thanks, done in https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/127/commits/211a4611a60d652e9e52d7e234521dbe2a767497

torch_np/_funcs_impl.py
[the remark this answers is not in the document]
lezcano: We would just need to import things

torch_np/_funcs_impl.py
```diff
@@ -1208,6 +1229,83 @@ def outer(a: ArrayLike, b: ArrayLike, out: Optional[OutArray] = None):
...
+
+    # work around 'bmm' not implemented for 'Half' etc
+    is_half = target_dtype == torch.float16
+    if is_half:
+        target_dtype = torch.float32
+
+    is_short_int = target_dtype in [torch.uint8, torch.int8, torch.int16, torch.int32]
+    if is_short_int:
+        target_dtype, result_dtype = torch.int64, target_dtype
```
lezcano, line 1279: `result_dtype` is not being used.

ev-br: thanks, removed.

NB it is not used because the current implementation follows pytorch in that e.g.

```
In [48]: a = np.arange(8, dtype=np.int8)

In [49]: torch.einsum(torch.as_tensor(a), [0], []).dtype
Out[49]: torch.int64
```

unlike numpy where the last line is `int8` (sigh)

lezcano: Yeah, this is because, as `sum` and reductions that accumulate, integeras are upcasted to `int64`

lezcano (approved):
Great! What a pain of a function...

ev-br:
Thanks Mario for the review!

PR #129, opened 2023-04-26
Fixes and minors

lezcano:
Minor improvements here and there e.g.
- prefer flatten over ravel as it's more PyTorch-y
- Prefer `.double()` or `.long()` over `to(float)` (I didn't even know that worked...) for the same reason
- Don't use out= explicitly, 
- Improved a bit the implementation of `average`.
- Don't call .item() if we can avoid it (added to the list of differences)
- remove _mappings
- remove the need for semi_private methods
- Fixed ndarray.fill, as the normalizer was not working because of the future annotations


torch_np/_reductions.py
```diff
@@ -326,13 +261,13 @@ def average_weights(a, axis, w, keepdims=False):
...
-    numerator = torch.mul(a, w).sum(axis)
-    denominator = w.sum(axis)
+    numerator = torch.mul(a, w).sum(axis, dtype=result_dtype)
+    denominator = w.sum(axis, dtype=result_dtype)
     result = numerator / denominator
 
     # keepdims
     if keepdims:
-        result = _util.apply_keepdims(result, ax, ndim)
+        result = _util.apply_keepdims(result, axis, a.ndim)
```
ev-br, line 270: IIRC this was problematic in of earlier iterations. As long as CI is green, it's OK now I suppose.

torch_np/_util.py
```diff
@@ -119,18 +119,17 @@ def apply_keepdims(tensor, axis, ndim):
...
         shape = expand_shape(tensor.shape, axis)
         tensor = tensor.reshape(shape)
     return tensor
 
 
-def axis_none_ravel(*tensors, axis=None):
-    """Ravel the arrays if axis is none."""
-    # XXX: is only used at `concatenate` and cumsum/cumprod. Inline unless reused more widely
+def axis_none_flatten(*tensors, axis=None):
+    """Flatten the arrays if axis is None else normalize the axis."""
```
ev-br, line 130: Nit does not normalize the axis.

ev-br (approved):

ev-br:
> Don't call .item() if we can avoid it (added to the list of differences)

I don't see it in https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/73, is this where you meant to add it ?

torch_np/_util.py
lezcano: Sorry, I was trying something else here but I didn't like it in the end. Forgot to update the comment

lezcano:
Yes, I added it, but after some editing I guess it was deleted along with the line about lists and tuples. This is fixed now.

PR #130, opened 2023-04-29
MAINT: add e2e example on CUDA

ev-br:
run examples in `e2e` folder on CUDA using the global  `set_default_device` switch.

In a nutshell, it just works, at least in the eager mode.


lezcano (approved):
This is great! I'll add this point to the RFC.
Let's just put a global flag `DEVICE="cpu"/"cuda"` and document this in the README

ev-br:
Global flag: you mean examples or `torch_np`? 
For examples I'm not sure it's any better than a comment really. For the wrappers, again not sure how that is meant to work without propagating `device=...`  arguments to all functions.

lezcano:
Yes, just for the examples. But yeah, then no global flag, but set `"cpu"` as the default.

e2e/mandelbrot/mandelbrot.py
```diff
@@ -6,6 +6,14 @@
 #import numpy as np
 import torch_np as np
 
+
+# To run on CUDA, change "cpu" to "cuda" below.
+# NB: set_default_device("cpu") leads to fails if cuda.is_available() is False, hence guard it.
```
lezcano, line 11: I can't repro that on my laptop
image

If you can repro it, can you report it as a bug?

ev-br: Nope, was a user fluke, is gone now.

PR #132, opened 2023-05-02
Wrap from_dlpack and its dunders

ev-br:
Just delegate to whatever self.tensor does under the hood.

- vendor the numpy tests suite
- remove datetimes, structured dtypes etc
- had to xfail refcounting related tests --- checked that pytorch does not pass these either.
- had to xfail a test with a non-None dlpack stream --- checked that pytorch does not raise

Also added a small test to check that tnp arrays can be created from / exported to pytorch tensors.

Poking around in ipython, it seems that `np.from_dlpack(torch.Tensor)` creates a readonly array, while the other direction it's a view. Not sure if there's something to do or if we can do anything about it really.

```
In [48]: a = np.arange(3)

In [49]: a
Out[49]: array([0, 1, 2])

In [50]: t = torch.from_dlpack(a)

In [51]: t[0] = 42

In [52]: t
Out[52]: tensor([42,  1,  2])

In [53]: a
Out[53]: array([42,  1,  2])
```

### DIFF
--- a/torch_np/_ndarray.py
+++ b/torch_np/_ndarray.py
@@ -442,6 +442,12 @@ class ndarray:
     take = _funcs.take
     put = _funcs.put
 
+    def __dlpack__(self, *, stream=None):
+        return self.tensor.__dlpack__(stream=stream)
+
+    def __dlpack_device__(self):
+        return self.tensor.__dlpack_device__()
+
 
 # This is the ideally the only place which talks to ndarray directly.
 # The rest goes through asarray (preferred) or array.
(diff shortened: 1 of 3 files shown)


torch_np/_ndarray.py
```diff
@@ -488,6 +494,11 @@ def asarray(a, dtype=None, order="K", *, like=None):
     return array(a, dtype=dtype, order=order, like=like, copy=False, ndmin=0)
 
 
+def from_dlpack(x, /):
+    t = torch.from_dlpack(x)
+    return ndarray(t)
+
+
```
lezcano, line 501: Why not in `_funcs_impl`?

lezcano (approved):
We don't have readonly tensors in PyTorch, so nothing to do there.

This LGTM, but I'll let @rgommers comment on this one, as he knows these functions better.

rgommers (approved):
This LGTM too, thanks @ev-br and @lezcano. Agreed that there's nothing to do regarding view vs. readonly array. Users are able to shoot themselves in the foot when they do in-place operations on tensors obtained from numpy arrays, but that's not specific to DLPack but rather a generic trait of operating on data in memory shared between two different libraries.

PR #134, opened 2023-05-08
ENH: add np.convolve and np.correlate

ev-br:


ev-br:
Have added `np.correlate` while at it @lezcano

lezcano (approved):

PR #135, opened 2023-05-10
ENH: control a random stream to be from either numpy or pytorch

ev-br:
Add a global switch to select the random stream origin: pytorch or numpy. If points to NumPy, `tnp.random.uniform` just calls `numpy.random.uniform` etc.

### DIFF
--- a/torch_np/random.py
+++ b/torch_np/random.py
@@ -7,6 +7,7 @@ Q: default dtype is float64 in numpy
 """
 from __future__ import annotations
 
+import functools
 from math import sqrt
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


torch_np/random.py
```diff
@@ -29,21 +30,44 @@
     "randint",
     "shuffle",
     "uniform",
+    "RANDOM_STREAM",
 ]
 
 
+RANDOM_STREAM = "pytorch"
```
lezcano, line 37: Just make it a boolean variable. Something like `USE_NUMPY_RANDOM = False`

lezcano (approved):
Just one small comment, otherwise it LGTM

lezcano:
You can also now use this flag in the E2E tests.

torch_np/random.py
ev-br: The string value here parallels `set_default_dtype`: https://github.com/Quansight-Labs/numpy_pytorch_interop/blob/main/torch_np/_dtypes.py#L346

lezcano: I still think it's better to have a boolean flag for this, as we just have two values. In the other case we also have other possible values, so the enum is alright. Although I'm not the biggest fan of the enum sugar in the dtypes either, but that's a story for another day ajjaja

ev-br: OK, will change. My experience is that almost all boolean flags (modulo a set of measure zero, that is) have to be changed to more than two state eventually. But here it's OK for now indeed, so let me change it and move on :-).

PR #137, opened 2023-05-12
NEP 50 scalars, weakly typed

ev-br:
cross-ref https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/136#issue-1704384620

Make python scalars weakly typed: `(np.uint8(1) + 2).dtype` is `int64` on main, and  `uint8` in this PR.

The price is a special flag `is_weak` we attach to _tensors_. Would be nicer to be able to attach them to dtypes, but torch.dtypes do not allow that.
The flag needs to be plumbed all the way from the `normalizer` decorator: otherwise torch-level implementer functions cannot tell if the numpy-level functions were called with `array(2)` (strongly typed) and scalar 2 (weakly typed). 

Currently, NEP 50 examples mostly pass, but there are several other test failures, so it's WIP for now.

PR #138, opened 2023-05-13
BUG: make array-detection logic in array(...) more robust

ev-br:
Needed for arrays hiding as elements of nested lists e.g. `np.array([[1, 2], [3, np.array(4)]])`.

Split off from gh-137.

### DIFF
--- a/torch_np/_ndarray.py
+++ b/torch_np/_ndarray.py
@@ -449,6 +449,19 @@ class ndarray:
         return self.tensor.__dlpack_device__()
 
 
+def _tolist(obj):
+    """Recusrively convert tensors into lists."""
+    a1 = []
+    for elem in obj:
+        if isinstance(elem, (list, tuple)):
+            elem = _tolist(elem)
+        if isinstance(elem, ndarray):
+            a1.append(elem.tensor.tolist())
+        else:
+            a1.append(elem)
+    return a1
+
+
 # This is the ideally the only place which talks to ndarray directly.
 # The rest goes through asarray (preferred) or array.
 
@@ -469,13 +482,7 @@ def array(obj, dtype=None, *, copy=True, order="K", subok=False, ndmin=0, like=N
 
     # lists of ndarrays: [1, [2, 3], ndarray(4)] convert to lists of lists
     if isinstance(obj, (list, tuple)):
-        a1 = []
-        for elem in obj:
-            if isinstance(elem, ndarray):
-                a1.append(elem.tensor.tolist())
-            else:
-                a1.append(elem)
-        obj = a1
+        obj = _tolist(obj)
 
     # is obj an ndarray already?
     if isinstance(obj, ndarray):
--- a/torch_np/_ndarray.py
+++ b/torch_np/_ndarray.py
@@ -467,7 +467,10 @@ def _tolist(obj):
 
 
 def array(obj, dtype=None, *, copy=True, order="K", subok=False, ndmin=0, like=None):
-    _util.subok_not_ok(like, subok)
+    if subok is not False:
+        raise NotImplementedError(f"'subok' parameter is not supported.")
+    if like is not None:
+        raise NotImplementedError(f"'like' parameter is not supported.")
     if order != "K":
         raise NotImplementedError
 
--- a/torch_np/_util.py
+++ b/torch_np/_util.py
@@ -19,13 +19,6 @@ def is_sequence(seq):
     return True
 
 
-def subok_not_ok(like=None, subok=False):
-    if like is not None:
-        raise ValueError("like=... parameter is not supported.")
-    if subok:
-        raise ValueError("subok parameter is not supported.")
-
-
 class AxisError(ValueError, IndexError):
     pass
 
--- a/torch_np/tests/test_basic.py
+++ b/torch_np/tests/test_basic.py
@@ -553,3 +553,9 @@ class TestExport:
         )
         diff = set(exported_fns).difference(set(dir(_np)))
         assert len(diff) == 0
+
+
+class TestCtorNested:
+    def test_arrays_in_lists(self):
+        lst = [[1, 2], [3, w.array(4)]]
+        assert_equal(w.asarray(lst), [[1, 2], [3, 4]])


torch_np/_ndarray.py
```diff
@@ -449,12 +449,28 @@ def __dlpack_device__(self):
         return self.tensor.__dlpack_device__()
 
 
+def _tolist(obj):
+    """Recusrively convert tensors into lists."""
+    a1 = []
```
lezcano, line 454: Bring here the if obj is isntance list/tuple, otherwise this function assumes that obj is an iterable, whcih is a non-trivial thing to assume in this context.

torch_np/_ndarray.py
```diff
@@ -469,13 +485,7 @@ def array(obj, dtype=None, *, copy=True, order="K", subok=False, ndmin=0, like=N
...
     # lists of ndarrays: [1, [2, 3], ndarray(4)] convert to lists of lists
     if isinstance(obj, (list, tuple)):
-        a1 = []
-        for elem in obj:
-            if isinstance(elem, ndarray):
-                a1.append(elem.tensor.tolist())
-            else:
-                a1.append(elem)
-        obj = a1
+        obj = _tolist(obj)
```
lezcano, line 488: Also, in what context do we need this? `as_tensor` already seems to do this:
```python
>>>torch.as_tensor([[2,3], np.array([2,3])])
tensor([[2, 3],
        [2, 3]])
```

ev-br: In main, it only looks into the first level of nesting:

```
In [3]: tnp.array([[1, 2], [tnp.int8(1), 2]])
IndexError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 tnp.array([[1, 2], [tnp.int8(1), 2]])

File ~/sweethome/proj/scipy/torch_np_compat/torch_np/_ndarray.py:495, in array(obj, dtype, copy, order, subok, ndmin, like)
    493 def array(obj, dtype=None, *, copy=True, order="K", subok=False, ndmin=0, like=None):
    494     # The result of the public `np.array(obj)` is not weakly typed.
--> 495     return _array(obj, dtype, copy=copy, order=order, subok=subok, ndmin=ndmin, like=like, is_weak=False)

File ~/sweethome/proj/scipy/torch_np_compat/torch_np/_ndarray.py:489, in _array(obj, dtype, copy, order, subok, ndmin, like, is_weak)
    486 if dtype is not None:
    487     torch_dtype = _dtypes.dtype(dtype).torch_dtype
--> 489 tensor = _util._coerce_to_tensor(obj, torch_dtype, copy, ndmin, is_weak)
    490 return ndarray(tensor)

File ~/sweethome/proj/scipy/torch_np_compat/torch_np/_util.py:213, in _coerce_to_tensor(obj, dtype, copy, ndmin, is_weak)
    211     tensor = torch.as_tensor(obj, dtype=dtype)
    212 else:
--> 213     tensor = torch.as_tensor(obj)
    215     # tensor.dtype is the pytorch default, typically float32. If obj's elements
    216     # are not exactly representable in float32, we've lost precision:
    217     # >>> torch.as_tensor(1e12).item() - 1e12
   (...)
    220     # Therefore, we treat `tensor.dtype` as a hint, and convert the
    221     # original object *again*, this time with an explicit dtype.
    222     torch_dtype = _dtypes_impl.get_default_dtype_for(tensor.dtype)

File ~/sweethome/proj/scipy/torch_np_compat/torch_np/_ndarray.py:391, in ndarray.__len__(self)
    390 def __len__(self):
--> 391     return self.tensor.shape[0]

IndexError: tuple index out of range
```

A smoke test is https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/138/files#diff-66e581e6a3a373197465b30dd84684d915c83f3c619773382e8f8a40140662a5R558

torch_np/_ndarray.py
ev-br: It's guarded with an if isinstance check at the call site?
https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/138/files#diff-83812bd01681b687b8fced4eac464b076aecffc471e0364d905b38fba8794e6cR487

torch_np/_ndarray.py
lezcano: But I think this works in PyTorch?
```python
>>> torch.as_tensor([[1,2],[np.int8(1), 2]])
tensor([[1, 2],
        [1, 2]])
```

ev-br: With numpy ndarrays, yes. With torch_np wrapper ndarrays, not in main yes with this PR.

ev-br:
Note that this function is *almost* identical to `_helpers/ndarrays_to_tensors`, which is used for fancy indexing and basically converts nested arrays to tensors. Letting go of `.tolist` here sounded tempting but does not quite work because numpy allows, for instance,

```
(Pdb) import numpy as _np
(Pdb) p x
array([[1., 2.],
       [3., 4.]], dtype=float32)
(Pdb) _np.asarray([x, 2*x, 3*x])
array([[[ 1.,  2.],
        [ 3.,  4.]],

       [[ 2.,  4.],
        [ 6.,  8.]],

       [[ 3.,  6.],
        [ 9., 12.]]], dtype=float32)

(Pdb) import torch
(Pdb) t
tensor([[1., 2.],
        [3., 4.]])
(Pdb) torch.as_tensor([t, 2*t, 3*t])
*** ValueError: only one element tensors can be converted to Python scalars
```

torch_np/_ndarray.py
lezcano: Yeah. What I'm saying is that the logic of this function assumes that if implicitly. The if instance should be inside this function, not in the caller site..

torch_np/_ndarray.py
lezcano: nvm, I just read https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/138#issuecomment-1546964624. Can you add a comment explaining this point?

torch_np/_ndarray.py
```diff
@@ -449,12 +449,28 @@ def __dlpack_device__(self):
         return self.tensor.__dlpack_device__()
 
 
+def _tolist(obj):
+    """Recusrively convert tensors into lists."""
```
lezcano, line 453: nit Recursively

lezcano (approved):
Just missing a couple nits discussed in the comments, but otherwise LGTM.

PR #139, opened 2023-05-13
MAINT: rework result_type_impl to accept *tensors, not dtypes

ev-br:
Make result_type inference use tensors internally. This simplifies internal usage a bit, and serves as a prerequisite for introducing weak scalars, cf gh-137 (off which this PR is split).

### DIFF
--- a/torch_np/_util.py
+++ b/torch_np/_util.py
@@ -135,37 +135,40 @@ def axis_none_flatten(*tensors, axis=None):
         return tensors, axis
 
 
-def typecast_tensors(tensors, target_dtype, casting):
-    """Dtype-cast tensors to target_dtype.
+def typecast_tensor(t, target_dtype, casting):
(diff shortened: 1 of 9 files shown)


torch_np/_util.py
```diff
@@ -193,7 +196,6 @@ def _coerce_to_tensor(obj, dtype=None, copy=False, ndmin=0):
     """
     if isinstance(obj, torch.Tensor):
         tensor = obj
-        base = None
     else:
         tensor = torch.as_tensor(obj)
         base = None
```
lezcano, line 201: also this one, now that we're at it. Here or in any other PR really.

lezcano (approved):
Cool!

torch_np/_util.py
ev-br: Yup, included in the next one.

PR #141, opened 2023-05-15
MAINT: simplify testing.assert_*, delegate to torch.assert_close

ev-br:
While at it, remove a bunch of dead code in testing/


torch_np/testing/utils.py
```diff
@@ -207,924 +124,30 @@ def assert_equal(actual, desired, err_msg="", verbose=True):
...
-              instead of this function for more consistent floating point
-              comparisons.
+    from torch_np._normalizations import normalize_array_like
 
-    Given two numbers, check that they are approximately equal.
-    Approximately equal is defined as the number of significant digits
-    that agree.
+    actual = normalize_array_like(actual)
+    desired = normalize_array_like(desired)
+    return (actual == desired).all()
```
lezcano, line 132: Is it expected that `assert_*` functions don't actually assert anything but just return a bool? That seems a bit weird to me.

lezcano: Also, is it alright to return false for nans?

lezcano (commented):
What's the reason for this big refactor? I'm a bit scared that we may introduce some bug in the testing facilities, which wouldn't be great.

ev-br:
Per offline discussion with @lezcano , let's shelve this PR.

PR #143, opened 2023-05-18
ENH: NEP 50 "weak scalars" with dtype and PyTorch defaults

lezcano:
Note that this PR is against the branch from https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/140.

This PR tries to solve in a clean way some simpler cases first (dtype is not None, or we have two tensors) to be able to deal with the tricky part having some preconditions on the data. This fixes a couple bugs that were latent in the other branch. It also adds a test to test for this case.

The logic to solve NEP 50 is still that of https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/140 for the tricky case.

It also tries to avoid iterations over lists that can be very slow. See if the timings of CI agree with this.

### DIFF
--- a/torch_np/_dtypes_impl.py
+++ b/torch_np/_dtypes_impl.py
@@ -65,77 +65,74 @@ def _dtype_for_scalar(py_type):
     }[py_type]
 
 
-categories = [
-    (torch.bool,),
-    (torch.uint8, torch.int8, torch.int16, torch.int32, torch.int64),
-    (torch.float16, torch.float32, torch.float64),
-    (torch.complex64, torch.complex128),
-]
-
-
-def category(dtyp):
-    for j, cat in enumerate(categories):
-        if dtyp in cat:
-            return j
-    raise ValueError(f"unknown dtype {dtyp}")
-
-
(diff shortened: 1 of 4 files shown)


ev-br:
The error on CI is that a test is bad under NEP50:

```
In [24]: np._set_promotion_state('weak')

In [25]: np.add(1.0, 1e-15, dtype=bool)
Out[25]: True

In [26]: np._set_promotion_state('legacy')

In [27]: np.add(1.0, 1e-15, dtype=bool)
Cell In[27], line 1
----> 1 np.add(1.0, 1e-15, dtype=bool)

UFuncTypeError: Cannot cast ufunc 'add' input 0 from dtype('float64') to dtype('bool') with casting rule 'same_kind'

```

lezcano:
This is so annoying. The uniform treatment of the `dtype` arg before doing anything else was so clean. ugh.

lezcano:
Ah, nevermind! The test passes under NEP 50 lol

lezcano:
removed the test. Also, CI still takes 20 min... Well, I tried.

ev-br:
Thanks Mario.

PR #145, opened 2023-05-24
ENH: add cross

ev-br:
vendor the implementation from numpy 1.24.x series (https://github.com/numpy/numpy/blob/v1.24.0/numpy/core/numeric.py#L1486-L1685)
 
- dispatching to `torch.linalg.cross` is cumbersome, numpy's version allows way more weird cases which torch.linalg.cross does not (`a.shape=(2, 1, 3), b.shape=(5, 3)` etc)
- numpy's version goes to some length to do internal computations inplace; here we avoid `out=...` for differentiability, and rely on the compiler to transform the code (or not --- squeezing the last drop of performance here does not seem to be top prio)


torch_np/_funcs_impl.py
```diff
@@ -1249,6 +1249,78 @@ def outer(a: ArrayLike, b: ArrayLike, out: Optional[OutArray] = None):
...
+            cp0[...] = a1 * b2 - a2 * b1
+            cp1[...] = a2 * b0 - a0 * b2
+            cp2[...] = a0 * b1 - a1 * b0
+        else:
+            assert b.shape[-1] == 2
+            cp0[...] = -a2 * b1
+            cp1[...] = a2 * b0
+            cp2[...] = a0 * b1 - a1 * b0
+
+    return moveaxis(cp, -1, axisc)
```
lezcano, line 1321: nit. `torch.moveaxis` as above for consistency?

lezcano (approved):
The OP explanation sounds about right. Thank you for this!

torch_np/_funcs_impl.py
ev-br: Yes, absolutely! Fixed.
The very fact that it worked (tensor -> ndarray via preprocessing in normalize -> ndarray via postprocessing in mormalize) can be in principle argued to be a minor bug. Not sure and definitely not worth spending effort on.

PR #146, opened 2023-05-24
pick some low-hanging fruit in advanced indexing

ev-br:
The main change here is to allow a list (here, `[2, 3, 4]`) in the right-hand side of 

```     
a = np.zeros((5, 5))
a[[[0], [1], [2]], [0, 1, 2]] = [2, 3, 4]
```

### DIFF
--- a/torch_np/_ndarray.py
+++ b/torch_np/_ndarray.py
@@ -437,7 +437,12 @@ class ndarray:
 
         if isinstance(value, torch.Tensor):
             value = _util.cast_if_needed(value, self.tensor.dtype)
-        return self.tensor.__setitem__(index, value)
+
+        try:
+            return self.tensor.__setitem__(index, value)
+        except TypeError:
+            value = torch.as_tensor(value, dtype=self.tensor.dtype)
+            return self.tensor.__setitem__(index, value)
 
     take = _funcs.take
     put = _funcs.put
--- a/torch_np/_ndarray.py
+++ b/torch_np/_ndarray.py
@@ -439,8 +439,9 @@ class ndarray:
         index = _helpers.ndarrays_to_tensors(index)
         index = ndarray._upcast_int_indices(index)
 
-        value = normalize_array_like(value)
-        value = _util.cast_if_needed(value, self.tensor.dtype)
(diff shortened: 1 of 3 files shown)


lezcano (approved):
Great. I like the optimisation for the hot path.

torch_np/_ndarray.py
```diff
@@ -437,7 +437,12 @@ def __setitem__(self, index, value):
 
         if isinstance(value, torch.Tensor):
             value = _util.cast_if_needed(value, self.tensor.dtype)
-        return self.tensor.__setitem__(index, value)
+
+        try:
+            return self.tensor.__setitem__(index, value)
+        except TypeError:
+            value = torch.as_tensor(value, dtype=self.tensor.dtype)
+            return self.tensor.__setitem__(index, value)
```
lezcano, line 445: Actually, no, I don't understand this one. Aren't we calling `ndarrays_to_tensors` on values up there? I think that the right fix would:
- Not do anything if `value` is a scalar
- Call `torch.as_tensor` on `value` otherwise.

ev-br: This would introduce the check if value is a scalar for all arguments.

ndarrays_to_tensors is not relevant here, since the failing test has a list value.

torch_np/_ndarray.py
```diff
@@ -433,10 +438,10 @@ def __getitem__(self, index):
     def __setitem__(self, index, value):
         index = _helpers.ndarrays_to_tensors(index)
         index = ndarray._upcast_int_indices(index)
-        value = _helpers.ndarrays_to_tensors(value)
 
-        if isinstance(value, torch.Tensor):
-            value = _util.cast_if_needed(value, self.tensor.dtype)
+        value = normalize_array_like(value)
+        value = _util.cast_if_needed(value, self.tensor.dtype)
```
lezcano, line 443: As discussed, let's do this if `type(value) not in {bool,int,float,complex}`, to avoid wrapping scalars into tensors unnecessarily.

lezcano (approved):
Just a minor point, otherwise LGTM

lezcano (approved):
Just a minor point, otherwise LGTM

PR #147, opened 2023-05-24
Add `ascontiguousarray`, remove several expired xfails

ev-br:


torch_np/_ndarray.py
```diff
@@ -504,6 +504,13 @@ def asarray(a, dtype=None, order="K", *, like=None):
     return array(a, dtype=dtype, order=order, like=like, copy=False, ndmin=0)
 
 
+def ascontiguousarray(a, dtype=None, order="K", *, like=None):
+    arr = array(a, dtype=dtype, order=order, like=like, copy=False, ndmin=0)
+    if not arr.tensor.is_contiguous():
+        arr.tensor = arr.tensor.clone(memory_format=torch.contiguous_format)
```
lezcano, line 510:
```suggestion
        arr.tensor = arr.tensor.contiguous()
```

torch_np/_ndarray.py
```diff
@@ -504,6 +504,13 @@ def asarray(a, dtype=None, order="K", *, like=None):
     return array(a, dtype=dtype, order=order, like=like, copy=False, ndmin=0)
 
 
+def ascontiguousarray(a, dtype=None, order="K", *, like=None):
+    arr = array(a, dtype=dtype, order=order, like=like, copy=False, ndmin=0)
```
lezcano, line 508: Probably marginally better to call `asarray` here forwarding the call to `asarray` unless we plan to implement `order="C"` in terms of this function.

lezcano (approved):

torch_np/_ndarray.py
ev-br: ...and avoid an extra copy because `copy=True`. Done.

PR #148, opened 2023-05-24
Unxfail several tests

ev-br:
Looked through numpy-vendored tests, esp for xfails which mentioned fancy indexing. 
this was to follow up on https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/146


lezcano (approved):
The point of xfail was that we would get automatic signals from the xfails we can support as we add support for them, but I guess that this is not the case when we xfail a whole class? At any rate, thank you for the clean up!

PR #149, opened 2023-05-30
implement np.pad

ev-br:
Only `mode="constant"` for now, and only scalar fill_values. Broadcasted fill_values do not seem to be supported by `torch.nn.functional.pad`, which this function delegates to.

### DIFF
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -1814,9 +1814,7 @@ def pad(array: ArrayLike, pad_width: ArrayLike, mode="constant", **kwargs):
         typ = int
     value = typ(value)
 
-
     pad_width = torch.broadcast_to(pad_width, (array.ndim, 2))
     pad_width = torch.flip(pad_width, (0,)).flatten()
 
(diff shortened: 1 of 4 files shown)


torch_np/_funcs_impl.py
```diff
@@ -1798,3 +1798,23 @@ def min_scalar_type(a: ArrayLike, /):
...
+        raise NotImplementedError
+    value = kwargs.get("constant_values", 0)
+    if array.dtype.is_floating_point:
+        typ = float
+    elif array.dtype.is_complex:
+        typ = complex
+    elif array.dtype == torch.bool:
+        typ = bool
+    else:
+        typ = int
```
lezcano, line 1814: Don't we have a PyTorch dtype -> Python type function?

lezcano (approved):

torch_np/_funcs_impl.py
ev-br: Indeed we didn't. Will extract this into a helper, together with some more cleanup.

PR #153, opened 2023-06-01
ENH: add histogramdd

ev-br:


torch_np/_funcs_impl.py
```diff
@@ -1945,6 +1945,61 @@ def histogram(
...
+        bins = normalize_seq_array_like(bins)
+        bins_dtypes = [b.dtype for b in bins]
+        bins = [_util.cast_if_needed(b, sample.dtype) for b in bins]
+
+    if range is not None:
+        range = range.flatten().tolist()
+
+    if weights is not None:
+        # range=... is required : interleave min and max values per dimension
+        mm = sample.amin(0), sample.amax(0)
```
lezcano, line 1987: nit use `torch.aminmax`.

torch_np/_funcs_impl.py
```diff
@@ -1945,6 +1945,61 @@ def histogram(
...
+        range = torch.cat(mm).reshape(sample.ndim, -1).T.flatten()
```
lezcano, line 1988: I really don't understand this line or how it works for arrays of ndim > 2.

ev-br: A potentially confusing bit is that `sample.ndim` is always 2 [sample shape is `num_observations, num_dimensions`], would writing it `.reshape(2, -1)` help?

This line just puts min and max values of samples in to a (super-confusing) pytorch format:

_range (sequence of python:float) – Defines the leftmost and rightmost bin edges in each dimension._

So this line needs to interleave min and max values per dimension. Thus reshape into a 2D tensor, transpose and flatten.

It would be _way_ clearer to have them as a sequence of  (min, max) pairs per dimension, but it is what it is. Demo:
(dropping a breakpoint into https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/153/files#diff-4c6650ffbcd7df3273d37a47226ec0bd984a6dff87bd3f0cc3a7140d922c335bR635):

```
(Pdb) sample
tensor([[0.4035, 0.8427, 0.1566],
        [0.6767, 0.0554, 0.6229],
        [0.7081, 0.0033, 0.6746],
        [0.6155, 0.7949, 0.4788],
        [0.8838, 0.0882, 0.6917],
        [0.0389, 0.3628, 0.0786],
        [0.9376, 0.4940, 0.0153],
        [0.5947, 0.9733, 0.1309],
        [0.3025, 0.4650, 0.1510],
        [0.0945, 0.4090, 0.4954]], dtype=torch.float64)

(Pdb) mm = sample.aminmax(dim=0)
(Pdb) torch.cat((mm[0], mm[1]))
tensor([0.0389, 0.0033, 0.0153, 0.9376, 0.9733, 0.6917], dtype=torch.float64)
(Pdb) torch.cat((mm[0], mm[1])).reshape(2, -1)
tensor([[0.0389, 0.0033, 0.0153],
        [0.9376, 0.9733, 0.6917]], dtype=torch.float64)
(Pdb) torch.cat((mm[0], mm[1])).reshape(2, -1).T
tensor([[0.0389, 0.9376],
        [0.0033, 0.9733],
        [0.0153, 0.6917]], dtype=torch.float64)
(Pdb) torch.cat((mm[0], mm[1])).reshape(2, -1).T.flatten()
tensor([0.0389, 0.9376, 0.0033, 0.9733, 0.0153, 0.6917], dtype=torch.float64)
```

The bins output of 
```
h, b = torch.histogramdd(sample, bins, range, density=bool(density), **w_kwd)
```

seems to agree:

```
(Pdb) p [bb[0] for bb in b]
[tensor(0.0389, dtype=torch.float64), tensor(0.0033, dtype=torch.float64), tensor(0.0153, dtype=torch.float64)]
(Pdb) p [bb[-1] for bb in b]
[tensor(0.9376, dtype=torch.float64), tensor(0.9733, dtype=torch.float64), tensor(0.6917, dtype=torch.float64)]
```

torch_np/_funcs_impl.py
ev-br: Thanks! Done.
An unexpected twist is that keyword-only arguments look great, until `t.amin(0)` works but `t.aminmax(0)` doesn't :-).

ev-br:
Actually, there is a slight problem: the return value,  `tensor, (sequence of tensors)` does not get normalized back into arrays properly, and  tensors from the sequence leak to user. Am on it.

ev-br:
> Actually, there is a slight problem: the return value, tensor, (sequence of tensors) does not get normalized back into arrays properly, and tensors from the sequence leak to user. Am on it.

Done in https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/153/commits/dff29e5389b7a7bd00ac0c91ac0b96924c9b2ae9. Am going to merge if/when the CI comes back green.

PR #155, opened 2023-06-09
ENH: add np.choose

ev-br:
Use  https://github.com/pytorch/pytorch/issues/9407#issuecomment-1427907939 for a broacasting replacement of `gather`.

Also only `mode="raise"` for now.

### DIFF
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -997,13 +997,9 @@ def choose(
 
     # Use an analog of `gather(choices, 0, a)` which broadcasts `choices` vs `a`:
     # (taken from https://github.com/pytorch/pytorch/issues/9407#issuecomment-1427907939)
-
-    input_shape = choices.shape
-    n_dims = choices.ndim
-
     idx_list = [
-        torch.arange(dim).view((1,) * i + (dim,) + (1,) * (n_dims - i - 1))
-        for i, dim in enumerate(input_shape)
+        torch.arange(dim).view((1,) * i + (dim,) + (1,) * (choices.ndim - i - 1))
+        for i, dim in enumerate(choices.shape)
     ]
 
     idx_list[0] = a
--- a/torch_np/_funcs_impl.py
+++ b/torch_np/_funcs_impl.py
@@ -986,6 +986,29 @@ def put_along_axis(arr: ArrayLike, indices: ArrayLike, values: ArrayLike, axis):
     return None
 
 
+def choose(a: ArrayLike, choices: Sequence[ArrayLike], out: Optional[OutArray]=None, mode: NotImplementedType="raise"):
+
+    # from https://github.com/pytorch/pytorch/issues/9407#issuecomment-1427907939
+    def gather_broadcast(input, dim, index):
+        input_shape = input.shape
+        n_dims = len(input_shape)
+
+        idx_list = [
+            torch.arange(input_shape[i])[
+                (None,) * i + (...,) + (None,) * (n_dims - i - 1)
+            ]
+            for i in range(n_dims)
+        ]
+        idx_list[dim] = index
+
+        return input[idx_list]
+
+
+    choices = torch.stack(torch.broadcast_tensors(*choices))
+    return gather_broadcast(choices, 0, a).squeeze(0)
+
+
+
 # ### unique et al ###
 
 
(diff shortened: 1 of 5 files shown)


torch_np/_funcs_impl.py
```diff
@@ -986,6 +986,30 @@ def put_along_axis(arr: ArrayLike, indices: ArrayLike, values: ArrayLike, axis):
...
+    choices = torch.stack(torch.broadcast_tensors(*choices))
+
+    # Use an analog of `gather(choices, 0, a)` which broadcasts `choices` vs `a`:
+    # (taken from https://github.com/pytorch/pytorch/issues/9407#issuecomment-1427907939)
+
+    input_shape = choices.shape
+    n_dims = choices.ndim
+
+    idx_list = [
+        torch.arange(input_shape[i])[(None,) * i + (...,) + (None,) * (n_dims - i - 1)]
```
lezcano, line 1005: I don't understand what you're doing here really. Isn't all this indexing code equivalent to `choices[a]`?

ev-br: This is basically `torch.gather(choices, 0, a)` only allowing `choices` to be a mix of 1D and ND arrays.

lezcano: Isn't it possible then to use to use `gather` by performing the performing the broadcast manually via `unsqueeze` and `view_as(a)`?

ev-br: Maybe. I do not see an elegant way of inserting an ellipsis in between of Nones in a list of e.g. [Ellipsis, None, None], [None, Ellipsis, None], [None, None, Ellipsis]. Also for the varying lengths of `arange(input_shape[i])` though.

ev-br: Can do of course `idx_list = torch.arange(input_shape[i]).view( (1,)*i + (input_shape[i],) + (1,)*(n_dims - i -1) ) for i in range(n_dims)`, but I really don't know if this reads any better than what's here now.

lezcano: So, the construction of `a[None,...]` is very much from NumPy and not so much from PyTorch. In PyTorch we often use `unsqueeze` instead. From a PyTorch prespective, the option with a `view` is much more readable, could you go with that one?

p.s what threw me off in the first read was that the ellipsis was simply used as a `:`. In that case it could have been more clear to go with `slice(None)`. Now, the `None` vs `slice(None)` would have then become REALLY confusing, so  yeah...

lezcano (approved):
Just one readability point, but otherwise LGTM

torch_np/_funcs_impl.py
```diff
@@ -1002,8 +1002,8 @@ def choose(
     n_dims = choices.ndim
```
lezcano, line 1002: nit. now `n_dims` is not used, and you can directly use `choices.shape` below rather an aux variable.

torch_np/_funcs_impl.py
ev-br: OK, done.

Out of curiosity, what would the pytorch-y way of `unsqueeze`-ing multple axes? Instead of `a[(None,)*3, ...]`

And yeah, I too wish `:` were a separate syntax outside of the square brackets, to parallel `...`. None vs slice(None) vs slice(None, None, None) gets ugly quick

torch_np/_funcs_impl.py
ev-br: `torch.arange(dim).view((1,) * i + (dim,) + (1,) * (choices.ndim - i - 1)`, really? OK, done.

torch_np/_funcs_impl.py
```diff
@@ -986,6 +986,26 @@ def put_along_axis(arr: ArrayLike, indices: ArrayLike, values: ArrayLike, axis):
...
+    out: Optional[OutArray] = None,
+    mode: NotImplementedType = "raise",
+):
+    # First, broadcast elements of `choices`
+    choices = torch.stack(torch.broadcast_tensors(*choices))
+
+    # Use an analog of `gather(choices, 0, a)` which broadcasts `choices` vs `a`:
+    # (taken from https://github.com/pytorch/pytorch/issues/9407#issuecomment-1427907939)
+    idx_list = [
+        torch.arange(dim).view((1,) * i + (dim,) + (1,) * (choices.ndim - i - 1))
```
lezcano, line 1001: ah, I didn't realised it's also used there, I thought it was just used in the enumerate lol. Sorry for that.

PR #163, opened 2023-07-12
compile  e2e smoke and nn_from_scratch benchmarks

ev-br:
move the eager version to `e2e/smoke/eager`, remove MPL from both eager and compiled versions, work around `astype(int)` to avoid a graph break.


e2e/smoke/smoke_nompl.py
```diff
@@ -0,0 +1,68 @@
...
+import torch
+torch.set_default_device("cpu")
+
+import torch._dynamo.config as cfg
+cfg.numpy_ndarray_as_tensor = True
+
+@torch.compile
+def update_single(N, dens, u, v, u_prev, v_prev, visc, diff, dt):
+    vel_step(N, u, v, u_prev, v_prev, visc, dt)
+    dens_step(N, dens, dens_prev, u, v, diff, dt)
```
ev-br, line 61: As discussed, this only compiles the inner function to avoid dynamo unrolling the whole loop.

lezcano (approved):
Might be nice to remove the `*_2.py` files, as discussed. Otherwise LGTM.

lezcano:
I'd be interested in how well does inductor do here, as there are a fair amount of in-place ops. I'd hope it's able to perform all of them out-of-place and then just `copy_` the result into the original tensor at the end. in-place ops are just too annoying for compilers, so we functionalise the graph to be able to reason about it properly, and then we recover the in-place semantics by adding a bunch of in-place copies from the functionalised outputs into the original tensors at the end.

ev-br:
Removed smoke/*_2.py files, also added the nn_from_scratch example, with tweaks enough so that it at least compiles. Perf comparisons to follow in a follow-up.

lezcano:
Cool!

ev-br:
So, 
- mojo/mandelbrot benchmark compiles fine. To avoid linear traces, we only compile the inner loop; the speedup w.r.t. numpy is about x8 with float64, and about 14x with float32 (for whatever reason, float64 is not vectorized C++). This is with a default OpenMP (x4 on a laptop, x32 on a server).
- bellman-ford e2e example simply does not compile because there is a data-dependent loop.

lezcano:
All these serve as *great* examples of the limitations of dynamo that we should put in the blogpost. The thing is that PyTorch users don't tend to use PyTorch for these things, but NumPy users do, so we should make sure to advertise these.

ev-br:
Й

ev-br:
Merged this. The CFD simulator might warrant some more looking into, which will do in a sequel.

PR #164, opened 2023-07-15
e2e : add an inner op of the kmeans algorithm

ev-br:
Add a clustering primitive, essentially a single iteration of the k-means algorithm, from https://realpython.com/numpy-array-programming/#clustering-algorithms
This is of course two lines of numpy :-).

Observations:
- `np.linalg.norm` does not compile (https://github.com/pytorch/pytorch/issues/105269), so cook up an alternative;
- the generated code is reasonable. The fun thing is that dynamo compiles `np.argmin(...)` into an explicit for-loop in C++ (wow);
- the speedup vs numpy is large, x10-x50, depending on the input size
- for very large inputs, numpy  runs out of memory, but the `torch.compile`-ed code keeps working fine.

### DIFF
--- a/e2e/kmeans/kmeans.py
+++ b/e2e/kmeans/kmeans.py
@@ -1,0 +1,58 @@
+# k-means step, given data, `X` and centroids
+# https://realpython.com/numpy-array-programming/#clustering-algorithms
+import numpy as np
+import torch
+torch.set_default_device("cpu")
+import torch._dynamo.config as cfg
+cfg.numpy_ndarray_as_tensor = True
+
+
+# np.linalg.norm replacement (2-norm only), https://github.com/pytorch/pytorch/issues/105269
+def norm(a, axis):
+    s = (a.conj() * a).real
+    return np.sqrt(s.sum(axis=axis))
+
+
+#@torch.compile
+def get_labels(X, centroids) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


e2e/kmeans/kmeans.py
```diff
@@ -0,0 +1,58 @@
...
+import numpy as np
+import torch
+torch.set_default_device("cpu")
+import torch._dynamo.config as cfg
+cfg.numpy_ndarray_as_tensor = True
+
+
+# np.linalg.norm replacement (2-norm only), https://github.com/pytorch/pytorch/issues/105269
+def norm(a, axis):
+    s = (a.conj() * a).real
```
lezcano, line 12: We don't support complex numbers in inductor, so this could very well be `a ** 2`

lezcano (approved):

e2e/kmeans/kmeans.py
ev-br: so https://github.com/pytorch/pytorch/issues/105267 is a wontfix?

lezcano: yup

ev-br: https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/165 has an example of working around not having complex dtypes.

PR #169, opened 2023-08-01
fix ldexp casting + unary ufunc dtypes

ev-br:
the second commit is just trivial MAINT while I'm looking at this file

### DIFF
--- a/torch_np/_binary_ufuncs_impl.py
+++ b/torch_np/_binary_ufuncs_impl.py
@@ -23,7 +23,6 @@ from torch import (
     heaviside,
     hypot,
     lcm,
-    ldexp,
     less,
     less_equal,
     logaddexp,
--- a/torch_np/_binary_ufuncs_impl.py
+++ b/torch_np/_binary_ufuncs_impl.py
@@ -23,6 +23,7 @@ from torch import (
     heaviside,
     hypot,
     lcm,
+    ldexp,
     less,
     less_equal,
     logaddexp,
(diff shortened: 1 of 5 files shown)


torch_np/_ufuncs.py
```diff
@@ -146,6 +146,49 @@ def matmul(
...
+        if isinstance(x1, torch.Tensor):
+            x1 = _util.typecast_tensor(x1, dtype, casting)
+        else:
+            x1 = torch.as_tensor(x1, dtype=dtype)
+    else:
+        if not isinstance(x1, torch.Tensor):
+            x1 = torch.as_tensor(x1)
+        if _dtypes_impl._category(x1.dtype) < 2:
+            # cast integers to float64
+            x1 = x1.double()
```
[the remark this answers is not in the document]
ev-br: `.double()` is actually wrong because we allow for user control of  the default dtypes, so it needs to be `default_dtypes.float_dtype`. Ditto for unary ufuncs, and I've extracted the relevant three-liner into `_util` in the updated PR. Does that count as a reuse in your books? :-).

lezcano (approved):
Factoring things out helps finding subtle bugs!

torch_np/_binary_ufuncs_impl.py
```diff
@@ -23,7 +23,6 @@
     heaviside,
     hypot,
     lcm,
-    ldexp,
```
lezcano, line 26: Actually, I think this should stay for consistency with the others.

ev-br: These import are only used for looping over the functions to decorate generically. Since we're wrapping it manually, what is the point?

lezcano: Yeah, at the moment it's a noop, but it's interesting to have the torch -> torch functions there in case we want to use them in the future.

ev-br:
Merged, thanks for the review @lezcano !

